Issue #37, opened 2023-06-16
gpt-3.5-turbo-16k Support


benjamin-kirkbride:
Would love to see gpt-3.5-turbo-16k support, along with the 0613 models.

See: https://community.openai.com/t/gpt-3-5-turbo-0613-function-calling-16k-context-window-and-lower-prices/263263

benjamin-kirkbride:
Actually, the base models will be getting the 0613 treatment soon, so probably not worth doing that.

> On June 27th, the stable gpt-3.5-turbo will be automatically upgraded to this new version.

benjamin-kirkbride:
I see this is going to be covered by https://github.com/simonw/llm/issues/17, but depending on how long it will take to be done, would be cool to have this out sooner :)

cw:
Is this not already in place? See https://til.simonwillison.net/llms/larger-context-openai-models-llm

benjamin-kirkbride:
Oh okay awesome, didn't realize. Would be awesome for llm to automatically bump to the 16k model when necessary, but maybe that's out of scope :)

PR #38, opened 2023-06-16
add 16k models

benjamin-kirkbride:
also added a `3.5` alias

fixes https://github.com/simonw/llm/issues/37


simonw:
Thanks for this!

PR #47, opened 2023-06-17
Handle case when no templates are found

sderev:


simonw:
Could you add a test for this? It should go here, before the demo template YAML files are created: https://github.com/simonw/llm/blob/7e2a8355deec331699c4e9f5f397aacc4483e5ec/tests/test_templates.py#L36-L59

simonw:
I think the more Unix-y behavior here would be for it to not output anything at all if there are no templates - that way it won't risk confusing any tools that are written to process the output from the tool.

Similar to how `ls` outputs nothing if a directory is empty.

sderev:
It doesn't print anything anymore. I also added a test to ensure the error is caught.

simonw:
Perfect, thank you!

PR #57, opened 2023-06-17
Add a defaults parameter to templates.

cmungall:
This allows default values to be filled for templates.

For example:

```yaml
prompt: Summarize this text in the voice of $voice
defaults:
  voice: GlaDOS
```

When running without `-p` it will choose the default:

```bash
curl -s 'https://til.simonwillison.net/macos/imovie-slides-and-audio' | \
  strip-tags -m | llm -t summarize
```

But you can override the defaults with `-p`:

```bash
curl -s 'https://til.simonwillison.net/macos/imovie-slides-and-audio' | \
  strip-tags -m | llm -t summarize -p voice Yoda
```

I got this:

> Text, summarize in Yoda's voice, I will: "Hmm, young padawan. Summary of this text, you seek. Hmmm. ...

### DIFF
--- a/docs/templates.md
+++ b/docs/templates.md
@@ -131,6 +131,34 @@ I got this:
 
 > My previous test subject seemed to have learned something new about iMovie. They exported keynote slides as individual images [...] Quite impressive for a human.
 
+## Specifying default parameters
+
+You can also specify default values for parameters, using a `defaults:` key.
+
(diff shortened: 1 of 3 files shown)


simonw:
This is a really good idea. I love that your PR includes the tests and documentation.

PR #66, opened 2023-06-29
Add support for custom schemas

kmad:
OpenAI functions allow for the output of LLMs to adhere to a more structured format, such as JSON. This PR adds the ability to define "schemas" using Pydantic classes, as explored in the `openai_function_call` python package, which can be coupled with other tools to use LLMs to parse unstructured text.

This was largely inspired by combining insights from this [Twitter post](https://twitter.com/kcimc/status/1668789461780668416) with [Simon's useful toolsets](https://simonwillison.net/2023/May/18/cli-tools-for-llms/) for working with LLMs.


simonw:
Wow, this is a fascinating PR!

I've been thinking about this from a slightly different angle - I've been considering adding a `extract` command which allows you to extract data using a schema, in a way that can then be written into a SQLite database or dumped out as JSON.

Your schemas mechanism looks like it has a bunch of overlap with what I have planned. I'm going to sit on this for a few days but I may end up combining your approach with mine (and crediting you, obviously).

kmad:
Sounds great! This is admittedly a bit of a hacky proof of concept so there's definitely room for improvement. 

If useful, the [openai_function_call](https://github.com/jxnl/openai_function_call) repo has more examples/ concepts to consider (e.g., dataframes, etc.) for the database extract concept. I did notice a dependency on Python >= 3.9 which is derived from this module, so depending on your desire for portability we may need to rewrite certain components without that restriction to the extent possible.

Happy to help push this forward.

jxnl:
> Sounds great! This is admittedly a bit of a hacky proof of concept so there's definitely room for improvement.
> 
> If useful, the [openai_function_call](https://github.com/jxnl/openai_function_call) repo has more examples/ concepts to consider (e.g., dataframes, etc.) for the database extract concept. I did notice a dependency on Python >= 3.9 which is derived from this module, so depending on your desire for portability we may need to rewrite certain components without that restriction to the extent possible.
> 
> Happy to help push this forward.

1) i think what makes more sense is actually blending in [SQLModel](https://sqlmodel.tiangolo.com/) with OpenAISchema then we get one use case.

simonw:
I suggest implementing something like this as an LLM command-line plugin.

Issue #91, opened 2023-07-11
New logging schema for conversations and continue mode


simonw:
> To implement continue mode I'm going to need to persist these things to the database.
>
> Which means I need a whole new schema, since I'm switching to using ULID IDs as part of this work.

_Originally posted by @simonw in https://github.com/simonw/llm/issues/85#issuecomment-1630180132_

simonw:
Current schema: https://github.com/simonw/llm/blob/23eeb0f322328db32c4398d17fe7de9c622c1986/docs/logging.md#L78-L91

But I want to instead log conversations and responses. Here's what those look like at the code level right now:

https://github.com/simonw/llm/blob/23eeb0f322328db32c4398d17fe7de9c622c1986/llm/models.py#L47-L52

https://github.com/simonw/llm/blob/23eeb0f322328db32c4398d17fe7de9c622c1986/llm/models.py#L74-L89

I think I'm going to replace or remove the `LogMessage` class entirely:

https://github.com/simonw/llm/blob/23eeb0f322328db32c4398d17fe7de9c622c1986/llm/models.py#L32-L44

simonw:
I'm a bit nervous about these ULIDs, which look like this:
```pycon
>>> str(ulid.ULID()).lower()
'01h51r2j69dbj1qma2874bywvw'
>>> str(ulid.ULID()).lower()
'01h51r2jvwkx2d8t8dweynrkvb'
>>> str(ulid.ULID()).lower()
'01h51r2kay2hqj9ymw5fmckmhw'
```
(They are case-insensitive, I think lower-case is visually prettier.)

The downside of these is that they aren't things people can type, unlike `llm --chat 34 "continue prompt"`. People will have to copy-and-paste them.

But... I expect most CLI usage of the continue mode to use `llm --continue` instead, which just uses the most recent ID without you needing to specify it.

The benefit of them is that they're globally unique, like UUIDs - which is great news if you want to e.g. run prompts on your local machine and then upload them to a shared space later.

Since shared prompt libraries feel like a useful thing to support, I'm going to use ULIDs.

simonw:
Calling them "conversations" does also mean that the `llm --chat 34` option should perhaps be something a bit longer.

`llm --conversation 34`

Since `-c` is already taken by `--continue`, maybe I use a surprising shortcut letter like `-x`?

simonw:
I think the option is `--cid` - where C ID is short for Conversation ID.

simonw:
[...]

PR #93, opened 2023-07-11
Log and load conversations from new DB schema

simonw:
Refs:
- #91

TODO:

- [x] Update `llm logs` command to use the new tables
- [x] Update schema displayed in documentation
- [x] Get all tests passing against the new schema

### DIFF
--- a/docs/help.md
+++ b/docs/help.md
@@ -83,7 +83,7 @@ Options:
   --no-stream                  Do not stream output
   -n, --no-log                 Don't log to database
   -c, --continue               Continue the most recent conversation.
-  --chat INTEGER               Continue the conversation with the given chat ID.
+  --cid, --conversation TEXT   Continue the conversation with the given ID.
   --key TEXT                   API key to use
   --save TEXT                  Save prompt with this template name
   --help                       Show this message and exit.
--- a/docs/logging.md
+++ b/docs/logging.md
@@ -62,7 +62,7 @@ import sqlite_utils
 import re
 db = sqlite_utils.Database(memory=True)
 migrate(db)
-schema = db["logs"].schema
+schema = db["responses"].schema
 
 def cleanup_sql(sql):
     first_line = sql.split('(')[0]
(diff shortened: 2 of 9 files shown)


simonw:
Just had this very satisfying session using the `-c` option:

`sqlite-utils schema "$(llm logs path)" conversations responses | llm -m 4 -s 'Write a SQLite SQL query to get back responses ordered by id desc joined against conversations to bring in conversation_name and conversation_model'`

> Based on your given table structures, here is an SQL query that joins the responses and conversations tables, with ordering based on the responses id (in descending). It also retrieves the name and model from the conversations table.
> 
> ```sql
> SELECT r.*,
>        c.name AS conversation_name,
>        c.model AS conversation_model
> FROM responses r
> LEFT JOIN conversations c
> ON r.conversation_id = c.id
> ORDER BY r.id DESC
> ```
> 
> Note: Replace the `*` with the actual columns you want to select from the responses table.

`llm -c 'Just the SQL query and do not use *, list out all the column names'`

> Here's an updated query with all column names from the responses table listed out:
> 
> ```sql
> SELECT r.id,
>        r.model,
>        r.prompt,
>        r.system,
>        r.prompt_json,
>        r.options_json,
>        r.response,
>        r.response_json,
>        r.conversation_id,
>        r.duration_ms,
>        r.datetime_utc,
>        c.name AS conversation_name,
>        c.model AS conversation_model
> FROM responses r
> LEFT JOIN conversations c
> ON r.conversation_id = c.id
> ORDER BY r.id DESC
> ```
> 
> In this query, r and c are used as aliases for the responses and conversations tables, respectively. This makes the query more readable and easier to write.

`llm -c 'Why did you use a left join?'`

> I used a LEFT JOIN to ensure that all records from the "responses" table are returned, even if there is not a matching record in the "conversations" table. It's often used when you don't want missing matches to be excluding data from your result set.
> 
> In this case, a LEFT JOIN would ensure that we still get our response data even if for some reason a response doesn't have an associated conversation. 
> 
> But if every response is guaranteed to have an associated conversation, you could use an INNER JOIN instead to only get records that have matching conversation ids in both tables. 
> 
> Your intended results and the state of the data generally determine the type of join you will use.
>

Issue #128, opened 2023-07-24
ReadTheDocs builds are failing


simonw:
https://readthedocs.org/projects/llm/builds/21391357/

<img width="831" alt="image" src="https://github.com/simonw/llm/assets/9599/e4e1efd2-8523-46b2-9abc-0f62a5f5cb82">

https://docs.readthedocs.io/en/stable/config-file/v2.html

simonw:
I think this will work:
```yaml
version: 2

build:
  os: ubuntu-22.04
  tools:
    python: "3.11"

sphinx:
  configuration: docs/conf.py

formats:
   - pdf
   - epub

python:
   install:
   - requirements: docs/requirements.txt
```

simonw:
https://readthedocs.org/projects/llm/builds/21391435/ build for that PR:

<img width="776" alt="image" src="https://github.com/simonw/llm/assets/9599/ab53e55f-81b3-47c4-b6c8-6cfe53831d6b">

https://llm--129.org.readthedocs.build/en/129/ worked.

simonw:
Fixed.

PR #129, opened 2023-07-24
.readthedocs.yaml

simonw:
Refs #128


simonw:
https://llm--129.org.readthedocs.build/en/129/ worked.

PR #143, opened 2023-08-03
Add support for Azure, OpenAI, Palm, Anthropic, Cohere Models - using litellm

ishaan-jaff:
I'm the maintainer of litellm https://github.com/BerriAI/litellm - a simple & light package to call OpenAI, Azure, Cohere, Anthropic API Endpoints

This PR adds support for models from all the above mentioned providers

Here's a sample of how it's used:

```
from litellm import completion

## set ENV variables
# ENV variables can be set in .env file, too. Example in .env.example
os.environ["OPENAI_API_KEY"] = "openai key"
os.environ["COHERE_API_KEY"] = "cohere key"

messages = [{ "content": "Hello, how are you?","role": "user"}]

# openai call
response = completion(model="gpt-3.5-turbo", messages=messages)

# cohere call
response = completion("command-nightly", messages)

# anthropic call
response = completion(model="claude-instant-1", messages=messages)
```

### DIFF
--- a/llm/default_plugins/openai_models.py
+++ b/llm/default_plugins/openai_models.py
@@ -9,6 +9,7 @@ import requests
 from typing import List, Optional, Union
 import json
 import yaml
+import litellm
 
 
 @hookimpl
@@ -211,7 +212,7 @@ class Chat(Model):
             # openai client library requires one
             kwargs["api_key"] = "DUMMY_KEY"
         if stream:
-            completion = openai.ChatCompletion.create(
+            completion = litellm.completion(
                 model=self.model_name or self.model_id,
                 messages=messages,
                 stream=True,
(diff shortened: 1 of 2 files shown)


simonw:
This is better handled as an LLM plugin, see https://llm.datasette.io/en/stable/plugins/tutorial-model-plugin.html

cmungall:
For anyone ending up here looking to bridge `llm` and `litellm` (a great combination!), a plugin isn't strictly necessary for minimal functionality.

You can add something like this to `extra-openai-models.yaml`

```yaml
- model_name: ollama/mixtral
  model_id: litellm-mixtral
  api_base: "http://0.0.0.0:8000"
```

It would be great if the docs (on either llm or litellm side) had some tips here

A plugin would still be great as this would allow dynamic introspection of available models, easier switching of proxy URLs etc, rather than having to hardwire into the config yaml.

PR #147, opened 2023-08-04
Allow for use of Pydantic v1 as well as v2.

cmungall:
Use v1-style configuration. While this is officially
deprecated (https://docs.pydantic.dev/latest/usage/model_config/),
it is supported by v2. I would assume this support will remain
for a while, this will help ease the transition to v2.

Background: llm may be used in combination with other libraries that may directly or indirectly depend on pydantic v1, this PR is designed to give groups the ability to combine llm with their other libraries and have some breathing space to transition

### DIFF
--- a/llm/default_plugins/openai_models.py
+++ b/llm/default_plugins/openai_models.py
@@ -9,7 +9,7 @@ try:
     from pydantic import field_validator, Field
 except ImportError:
     from pydantic.fields import Field
-    from pydantic.class_validators import validator as field_validator # type: ignore [no-redef]
+    from pydantic.class_validators import validator as field_validator  # type: ignore [no-redef]
 import requests
 from typing import List, Optional, Union
 import json
--- a/llm/default_plugins/openai_models.py
+++ b/llm/default_plugins/openai_models.py
@@ -4,6 +4,7 @@ from llm.utils import dicts_to_table_string
 import click
 import datetime
 import openai
+
 try:
     from pydantic import field_validator, Field
 except ImportError:
--- a/llm/default_plugins/openai_models.py
+++ b/llm/default_plugins/openai_models.py
@@ -4,7 +4,11 @@ from llm.utils import dicts_to_table_string
 import click
 import datetime
 import openai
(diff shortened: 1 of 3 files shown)


cmungall:
@simonw any chance you could take a look at this? The rest of the ecosystem has yet to move over to pydantic v2, e.g. [chromadb is pinning to <2](https://github.com/chroma-core/chroma/pull/799), which means that currently chromadb and llm can't be used together...

simonw:
I've been bitten by this issue myself - https://github.com/simonw/llm-replicate almost couldn't use the official Replicate client library because it depended on Pydantic v1. I submitted a PR to help with that:

- https://github.com/replicate/replicate-python/pull/121

Then the `llm-claude` plugin had a similar issue:

- https://github.com/tomviner/llm-claude/issues/1

I think you're right: this needs to be fixed in LLM itself.

I want `llm` to be a useful dependency for other projects in [its Python library form](https://llm.datasette.io/en/stable/python-api.html) - but lots of other things remain on Pydantic 1.

Having it depend on just `pydantic` and work with both 1 and 2 would be ideal, at least for the next few months while the world upgrades to v2.

simonw:
One catch: LLM plugins will have to learn to not depend on Pydantic v1 v.s. v2 either. I'll need to document that in the plugin documentation I think.

Also I should set up CI tests that exercise both versions of Pydantic to make sure one of them doesn't accidentally break in the future.

cmungall:
Thanks!

I love pydantic, but version 2 is causing a lot of churn in my ecosystem right now.

tomviner:
Does look like supporting both is the best approach right now, during this crossover period. Libraries who've upgraded get asked for v1 compat (this PR), and those still on v1 get asked for v2 compat (https://github.com/anthropics/anthropic-sdk-python/issues/115).

Where's the `six` of `pydantic`?

tomviner:
Anthropic went cross compatible in https://github.com/anthropics/anthropic-sdk-python/pull/121 So I was able to unvendor them in the latest llm-claude release.

Issue #185, opened 2023-08-26
Initial abstraction for running embeddings


simonw:
I want to start experimenting more with Retrieval Augmented Generation. As part of that, I want to be able to calculate embeddings against different models.

I want `llm` to grow a `llm embed` command and a Python API for that too, which have their own plugin hook.

simonw:
Very early initial design idea. A new hook implementation:
```python
@hookspec
def register_embedding_models(register):
    "Return a list of model instances that can be used for embedding"
```
Each of those models will have a `.embed(text)` method that returns a list of floats.

Then `llm.get_embedding_model()` and `llm.get_embedding_models()` functions.

simonw:
Moving this to a PR:
- https://github.com/simonw/llm/pull/186

simonw:
The method currently returns a Python list of floats.

I know from experience that I usually end up wanting to store these as a packed array of binary floats (so I can put them in a SQLite blob) - usually using this:

```python
def decode(blob):
    return struct.unpack("f" * 1536, blob)


def encode(values):
    return struct.pack("f" * 1536, *values)
```
From https://github.com/simonw/openai-to-sqlite/blob/84d6bbd67379a054347dc9a7f84e6c7e225bfb67/openai_to_sqlite/cli.py#L362-L367

I should figure out where this bit fits into LLM itself.

simonw:
I think the CLI `embed` command defaults to returning a JSON array, but maybe that can return a comma-separated list of floats or even a binary blog based on some command options.

simonw:
A funny thing about embeddings is that they're only actually useful if you can make comparisons against them.

LLM already has a SQLite database for storing prompts. Maybe LLM itself also grows mechanisms for optionally storing embeddings in its own database, plus commands/functions for searching against them?

The problem there is that SQLite on its own will need to use brute-force for this, which will be slow.

But... if there was a plugin hook this could be a lot better. I could let people install plugins for FAISS and sqlite-vss and Pinecone and suchlike.

So out of the box LLM can give you slow vector search using brute force against embeddings stored in SQLite. But if you need faster lookups they should just be one plugin away.

simonw:
(I have a 5 hour plane ride today and my initial goal for this issue was to give myself the pieces I needed to hack on Retrieval Augmented Generation experiments on the plane... which I've now got, thanks to that prototype `llm-embed-sentence-transformers` plugin in https://github.com/simonw/llm/pull/186#issuecomment-1694485334

[...]

PR #186, opened 2023-08-26
Initial CLI support and plugin hook for embeddings

simonw:
Refs:
- #185

### DIFF
--- a/docs/embeddings/binary.md
+++ b/docs/embeddings/binary.md
@@ -1,0 +1,21 @@
+(embeddings-binary)=
+# Binary embedding formats
+
+The default output format of the `llm embed` command is a JSON array of floating point numbers.
+
+LLM stores embeddings in a more space-efficient format: little-endian binary sequences of 32-bit floating point numbers, each represented using 4 bytes.
+
+The following Python functions can be used to convert between the two formats:
+
+```python
+import struct
+
+def encode(values):
+    return struct.pack("<" + "f" * len(values), *values)
+
+def decode(binary):
+    return struct.unpack("<" + "f" * (len(binary) // 4), binary)
+```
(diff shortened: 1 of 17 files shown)


simonw:
Here's an example plugin that works with https://huggingface.co/sentence-transformers/all-MiniLM-L6-v2

```python
import llm
from sentence_transformers import SentenceTransformer


@llm.hookimpl
def register_embedding_models(register):
    model_id = "sentence-transformers/all-MiniLM-L6-v2"
    register(SentenceTransformerModel(
        model_id, model_id, 384)
    )


class SentenceTransformerModel(llm.EmbeddingModel):
    def __init__(self, model_id, model_name, embedding_size):
        self.model_id = model_id
        self.model_name = model_name
        self.embedding_size = embedding_size
        self._model = None

    def embed(self, text):
        if self._model is None:
            self._model = SentenceTransformer(self.model_name)
        return list(self._model.encode([text])[0])
```

simonw:
I introduced https://github.com/simonw/sqlite-migrate here, because I need a mechanism for ensuring any database passed to `llm embed -d/--database` has the necessary tables - independent of the migrations that create the `responses` etc tables in the main `logs.db` database.

I should absolutely port the other LLM migrations to this system too, having two migrations systems is gnarly.

simonw:
In order to write tests for this I'm going to build a plugin that only activates during tests, which has a VERY simple embedding mechanism: it returns 16 floats, each float is the length of a word in the input text, truncates to 16 words and fills in with `0` for missing words.

I'll use this pattern: https://til.simonwillison.net/pytest/registering-plugins-in-tests

simonw:
I'm going to document this, then land it and then work on `llm similar` separately.

Issue #253, opened 2023-09-09
Support for embedding binary files


simonw:
I want to be able to store and calculate embeddings for binary data - CLIP for images, [ImageBind](https://github.com/facebookresearch/ImageBind) for audio and suchlike.

`llm embed-multi` and `llm embed` currently assume text data, see:
- #225

Plus the embeddings models are expected to work against lists of strings.

I think a `--binary` flag in a few places plus redefining embedding models to optionally accept binary objects in addition to strings would be good.

simonw:
Here's a prototype which was enough to get a rough CLIP embedding model working:
```diff
diff --git a/llm/cli.py b/llm/cli.py
index d352087..c46a4d0 100644
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -1113,6 +1113,7 @@ def embed(collection, id, input, model, store, database, content, metadata, form
     help="Encoding to use when reading --files",
     multiple=True,
 )
+@click.option("--binary", is_flag=True, help="Treat --files as binary data")
 @click.option("--sql", help="Read input using this SQL query")
 @click.option(
     "--attach",
@@ -1135,6 +1136,7 @@ def embed_multi(
     format,
     files,
     encodings,
+    binary,
     sql,
     attach,
     prefix,
@@ -1158,6 +1160,10 @@ def embed_multi(
     2. A SQL query against a SQLite database
     3. A directory of files
     """
+    if binary and not files:
+        raise click.UsageError("--binary must be used with --files")
+    if binary and encodings:
+        raise click.UsageError("--binary cannot be used with --encoding")
     if not input_path and not sql and not files:
         raise click.UsageError("Either --sql or input path or --files is required")
 
@@ -1200,11 +1206,14 @@ def embed_multi(
                 for path in pathlib.Path(directory).glob(pattern):
                     relative = path.relative_to(directory)
                     content = None
-                    for encoding in encodings:
-                        try:
-                            content = path.read_text(encoding=encoding)
-                        except UnicodeDecodeError:
-                            continue
+                    if binary:
+                        content = path.read_bytes()
+                    else:
+                        for encoding in encodings:
+                            try:
+                                content = path.read_text(encoding=encoding)
+                            except UnicodeDecodeError:
+                                continue
                     if content is None:
                         # Log to stderr
                         click.echo(
@@ -1249,8 +1258,10 @@ def embed_multi(
             for row in rows:
                 values = list(row.values())
                 id = prefix + str(values[0])
[...]

PR #254, opened 2023-09-09
Binary embeddings

simonw:
Refs:
- #253

Still needed:

- [x] Decide what to do about `--store` - should it store the binary content in a new `content_blob` column? See [these notes](https://github.com/simonw/llm/issues/253#issuecomment-1712593336).
- [x] The `model.supports_binary` boolean needs to be respected - it should raise errors if you attempt to embed binary content against a text-only model
- [x] Should there be a way to mark a model as ONLY accepting binary content? Probably yes - though the models I have played with so far like CLIP and [ImageBind](https://github.com/facebookresearch/ImageBind) are happy to accept either kind of content.
- ~Lots of tests, using a mock embedding model that can handle binary content.~
- ~Build a plugin that uses this (I have a draft `llm-clip` one already).~


simonw:
I can build plugins using:

- https://www.sbert.net/examples/applications/image-search/README.html#usage
- https://github.com/mlfoundations/open_clip
- https://github.com/facebookresearch/ImageBind - looks like I'll need to vendor their code since it isn't on PyPI

simonw:
Here's my rough CLIP plugin:
```python
import llm
from PIL import Image
from sentence_transformers import SentenceTransformer, util
import io


@llm.hookimpl
def register_embedding_models(register):
    register(ClipEmbeddingModel())


class ClipEmbeddingModel(llm.EmbeddingModel):
    model_id = "clip"

    def __init__(self):
        self._model = None
        self._processor = None
        self._tokenizer = None

    def embed_batch(self, items):
        # Embeds a mix of text strings and binary images
        if self._model is None:
            self._model = SentenceTransformer('clip-ViT-B-32')

        to_embed = []

        for item in items:
            if isinstance(item, bytes):
                # If the item is a byte string, treat it as image data and convert to Image object
                to_embed.append(Image.open(io.BytesIO(item)))
            elif isinstance(item, str):
                to_embed.append(item)

        embeddings = self._model.encode(to_embed)
        return [[float(num) for num in embedding] for embedding in embeddings]
```

simonw:
I'm going to try the `content_blob` column and see how it feels.

simonw:
If a model only supports binary and does not support text, maybe we can have it treat all input as `--binary` even if you forget to use that flag?

simonw:
I'm landing this. Future tests will happen as I write the plugins.

PR #257, opened 2023-09-10
Updated error message for invalid or missing embedding model

almet:
Hey, just an improvement on the error message for missing default embedding model. I was confused with default models here (thought it was asking me for a LLM model)


simonw:
Thanks, I like this improvement.

PR #261, opened 2023-09-11
[Proof of concept] Support shell autocomplete for templates

pneff:
Together with click's Shell Completion support, this allows to auto-complete `llm -t <tab>` invocations.

This is an initial proof of concept. If you think this is the right thing to do, I can invest a little time to polish it. A few tasks I see as open:

* Check which other arguments would benefit from such treatment.
* Extend the documentation to include the autocomplete setup in the user guide.
* Fix auto-completion on the initial command. For example `llm <tab>` currently doesn't work.
* Performance is an issue, it's rather slow on my end.

For information how to set up the autocomplete here, see the official [Click Shell Completion documentation](https://click.palletsprojects.com/en/8.1.x/shell-completion/).

### DIFF
--- a/docs/help.md
+++ b/docs/help.md
@@ -90,7 +90,7 @@ Options:
   -s, --system TEXT            System prompt to use
   -m, --model TEXT             Model to use
   -o, --option <TEXT TEXT>...  key/value options for the model
-  -t, --template TEXT          Template to use
+  -t, --template TEMPLATE      Template to use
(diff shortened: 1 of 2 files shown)


pneff:
I have extended this slightly to auto-complete models as well. Template auto-complete also is fixed now to actually complete based on partial template names.

For command completion I sources the root cause, see click-contrib/click-default-group#22.

@simonw let me know if you are happy for me to continue with this PR. I would then add the documentation and some test coverage here.

PR #263, opened 2023-09-12
change type of cli "embed -c" to be STR not FILE

mhalle:
The type of the "content" argument was erroneously set to be FILE with a "type=click.Path(...)" argument, which content is a string.

### DIFF
--- a/docs/help.md
+++ b/docs/help.md
@@ -480,7 +480,7 @@ Options:
   -m, --model TEXT                Embedding model to use
   --store                         Store the text itself in the database
   -d, --database FILE
-  -c, --content FILE              Content to embed
+  -c, --content TEXT              Content to embed
   --metadata TEXT                 JSON object metadata to store
   -f, --format [json|blob|base64|hex]
                                   Output format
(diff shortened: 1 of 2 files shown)


simonw:
Huh! I'm surprised that didn't result in any test failures. Here's the commit with the mistake: https://github.com/simonw/llm/commit/77cf56e54a96040ba00dae18c6a8597977b9f7a9

simonw:
I think this is the test that exercised that: https://github.com/simonw/llm/blob/5ba34dbe360bda802116a2b36dc1d7d7c8618d4b/tests/test_embed_cli.py#L43-L58

simonw:
I figured out why it was mostly working - it's because this just said it had to be a writable path, and most strings are also writable paths:

https://github.com/simonw/llm/blob/5ba34dbe360bda802116a2b36dc1d7d7c8618d4b/llm/cli.py#L1039-L1044

But if I do this:
```bash
touch no-write
chmod -w no-write
llm embed -c no-write -m ada-002
```
I get this error:
```
Usage: llm embed [OPTIONS] [COLLECTION] [ID]
Try 'llm embed --help' for help.

Error: Invalid value for '-c' / '--content': File 'no-write' is not writable.
```
Which is fixed by this PR.

Issue #113, opened 2023-07-16
Structured Extraction w/ Function Calls


jxnl:
Hey I wrote a small [lib](https://openai-function-call.onrender.com/) around using function calls and pydatnic for extraction and i think some ideas could be appleid via the CLI or API
I really like the CLI interface so i'd love to brain some some ways of doing the extraction via CLI.

Heres an [example from my docs](https://openai-function-call.onrender.com/examples/search/)

The simplest is defining a model and making the model use it 

```python
class UserDetails(BaseModel):
    "Correctly extracted user information"
    name: str = Field(..., description="User's full name")
    age: int
```

Suggestion:
```
model("gpt-4", response_model=UserDetails)
```

In the template you could even support some pattern like:

```
model: gpt-4
system: Extract out a user detail
prompt: "look at {input} and ..."
response_model:
    - UserDetail
       description: ...
       parameters: 
           - name
             type: str 
            ...
```

But a more interesting use case could be configuring it to extract Multiple UserDetails

```
class UserDetails(BaseModel):
    "Correctly extracted user information"
    name: str = Field(..., description="User's full name")
    age: int
    
class MultipleDetails(BaseModel):
    batch: List[UserDetails]
```

If this was implemented well you could imagine having some cli tool out as jsonlines vs a single json object. It could even be streamed!

```
curl -s '...' | \
  strip-tags -m | llm --extract-multiple schema.json
```
returns 
```
{name: .., age: ...}
{name: .., age: ...}
```

cmungall:
I like this idea. It would be useful to have the ability to reference an existing pydantic model e.g. via package path, rather than making the user do a compilation step and copy the results into their templates

@simonw mentions the possibility of adding an `extract` command in #66

I'm also very in a command like this, I was thinking of implementing in a lib that depends on llm, but even better if it can be incorporated into `llm`.

I was thinking of implementing this in a way that can be made independent of openai/functions, and allow the user to choose between strategies:

1. Use openai functions, if available
2. just ask for the json/yaml directly, providing either in-context examples of a description of the schema
3. Using a recursive descent approach, as in our [SPIRES](https://arxiv.org/abs/2304.02711) algorithm, as implemented in our [OntoGPT](https://github.com/monarch-initiative/ontogpt) (this package does a lot more so the idea would be to scoop out the minimal core into a separate llm-extract package)

the second is highly error prone, but has the advantage of working with llama2 etc

Perhaps the last would be best suited to a plugin?

[...]

PR #281, opened 2023-09-16
Enable GPT functions

ihh:
Mentioned by @jxnl and @cmungall in #113. Actually this is a lot simpler/lighter than what is discussed there and in @kmad's PR #66, but it does what I need it to, and wraps the OpenAI API pretty simply I think

### DIFF
--- a/llm/default_plugins/openai_models.py
+++ b/llm/default_plugins/openai_models.py
@@ -293,8 +293,9 @@ class Chat(Model):
             kwargs["api_key"] = "DUMMY_KEY"
         if self.headers:
             kwargs["headers"] = self.headers
-        if kwargs.get("function_call") is not None:
-            kwargs["function_call"] = { "name": kwargs["function_call"] }
+        req_function_call = kwargs.get("function_call")
+        if req_function_call is not None:
+            kwargs["function_call"] = { "name": req_function_call }
         if stream:
             completion = openai.ChatCompletion.create(
                 model=self.model_name or self.model_id,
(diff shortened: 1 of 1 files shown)


simonw:
Initially I had thought that this wasn't the right direction to go - that function call support should be integrated more deeply into the library rather than just being provided by options like this.

But when I started thinking through that deeper integration, I realized that using options as a starting point was likely the first step on getting this to work!

I don't think the path-to-a-file thing is quite the right default though. I just tried switching the option to accepting JSON directly and got [this example](https://til.simonwillison.net/gpt3/openai-python-functions-data-extraction) to work:

```bash
llm -o functions '[
    {
        "name": "extract_locations",
        "description": "Extract all locations mentioned in the text",
        "parameters": {
            "type": "object",
            "properties": {
                "locations": {
                    "type": "array",
                    "items": {
                        "type": "object",
                        "properties": {
                            "name": {
                                "type": "string",
                                "description": "The name of the location"
                            },
                            "country_iso_alpha2": {
                                "type": "string",
                                "description": "The ISO alpha-2 code of the country where the location is situated"
                            }
                        },
                        "required": [
                            "name",
                            "country_iso_alpha2"
                        ]
                    }
                }
            },
            "required": [
                "locations"
            ]
        }
    }
]' 'I went to London and Utrecht last week'
```
Output:
```json
{"function_call": {
"name": "extract_locations",
"arguments": {
  "locations": [
    {
      "name": "London",
      "country_iso_alpha2": "GB"
    },
    {
      "name": "Utrecht",
      "country_iso_alpha2": "NL"
    }
  ]
}
}}
```
I'm going to land this as a starting point and continue to iterate on it (though it's going to be a bit tricky to fix the merge conflicts with the changes I made in #284)

jxnl:
what are your thoughts on being able to enable a path to a pydantic object?

where 
```python
# models.py
class ExtractLocation(BaseModel): 
    ...
```

```sh
llm -o model models.ExtractLocation 'I went to London and Utrecht last week'
```
```
{
  "locations": [
    {
      "name": "London",
      "country_iso_alpha2": "GB"
    },
    {
      "name": "Utrecht",
      "country_iso_alpha2": "NL"
    }
  ]
}
```

Theres a few workflows where i imagine i'd want to be able to 

`cat *.txt | llm -o model models.ExtractLocation  > data.jsonl`

simonw:
That's the direction I was initially thinking about going in, since LLM has Pydantic as a dependency already.

I'm still trying to flesh out all of the different ways this could be used. The thing I'm personally most interested in is being able to add a docstring and some type annotations to a Python function and have the functions schema for that generated automatically - but defining a Pydantic class (or just a regular Python dataclass) for the extract-structured-data case is interesting too.

simonw:
Part of the design challenge here is that I want to address all of these:

- The CLI use-case
- The interactive chat use-case (`llm chat` but also the web chat UI I'm putting together)
- The Python library use-case

jxnl:
Yeah, I think the advantages of pydantic really come in if you have more complex nested structures. I have some examples on my docs, but they would probably be too complicated to define in the command line. 

Often it's noy extract one object. But yield many objects from the text.

PR #310, opened 2023-10-24
Fix typo, command below references 7B model

admp:


mhaeussermann:
#305 includes the same fix.

admp:
@mhaeussermann thanks! closing.

PR #337, opened 2023-11-08
docs: add docs on how to configure models deployed via azure

bnookala:
I noted that openai models deployed to Azure were possible to utilize via `llm`, but could not find documentation on how to configure it. Thus, here it is 🎉 

This pull request adds documentation for using OpenAI models deployed via Azure OpenAI Service in `docs/other-models.md`. The most important change is the addition of instructions on how to add the model to the `extra-openai-models.yaml` file and an example of how to run a prompt against the added model using `llm`.

Main documentation changes:

* <a href="diffhunk://#diff-1c049db75a7dd8da8a8f1de60b272386d887069a862ea23d77cdce74f662f0eaR104-R136">`docs/other-models.md`</a>: Added documentation for using OpenAI models deployed via Azure OpenAI Service, including instructions on how to add the model to the `extra-openai-models.yaml` file and an example of how to run a prompt against the added model using `llm`.

### DIFF
--- a/docs/other-models.md
+++ b/docs/other-models.md
@@ -101,6 +101,33 @@ And confirm they were logged correctly with:
 llm logs -n 1
 ```
 
+### Azure OpenAI Service
+
+LLM supports OpenAI models deployed via [Azure OpenAI Service](https://azure.microsoft.com/en-us/products/ai-services/openai-service). These can also be added using the same configuration mechanism.
+
+The `model_id` is the name LLM will use for the model. The `model_name` is the deployment name of the model on Azure OpenAI Service. 
+
+The `api_base` will be the Azure OpenAI Service endpoint. The `api_key_name` is the name of the Azure OpenAI Service API key stored using the {ref}`api-keys` feature.
+
+You must add `api_version` to use a specific version of the Azure OpenAI Service API.
+
+Set `api_type` to `azure` to specify that the OpenAI client library should use the Azure OpenAI Service API.
+
+You must also add `api_engine` to specify the base model to use, eg: `gpt-35-turbo-16k`
+
+For example, to add the `mygpt35` model (a `gpt-35-turbo-16k` model) deployed on Azure OpenAI Service, using the `2023-05-15` version of the API, with the account endpoint `https://my-openai-account.openai.azure.com/`, add this to your `extra-openai-models.yaml` file:
+
+
+```yaml
+- model_id: mygpt
+  model_name: mygpt35
+  api_base: https://my-openai-account.openai.azure.com/
+  api_key_name: azure
+  api_version: 2023-05-15
+  api_type: azure
+  api_engine: gpt-35-turbo-16k
+```
+
 ### Extra HTTP headers
 
 Some providers such as [openrouter.ai](https://openrouter.ai/docs) may require the setting of additional HTTP headers. You can set those using the `headers:` key like this:


Rajesh-Veeranki:
Thanks for the documentation. I tried and it works. However, do you know if we can use the same approach for embedding models on Azure?

bnookala:
> Thanks for the documentation. I tried and it works. However, do you know if we can use the same approach for embedding models on Azure?

Thanks for verifying! I took a look at this, and I don't think this is possible without implementing a plugin to load custom embedding models. Might take an attempt at this over the weekend :)

docs/other-models.md
```diff
@@ -101,6 +101,33 @@ And confirm they were logged correctly with:
 llm logs -n 1
 ```
 
+### Azure OpenAI Service
+
+LLM supports OpenAI models deployed via [Azure OpenAI Service](https://azure.microsoft.com/en-us/products/ai-services/openai-service). These can also be added using the same configuration mechanism.
```
pamelafox, line 106:
```suggestion
LLM supports OpenAI models deployed via [Azure OpenAI Service](https://azure.microsoft.com/products/ai-services/openai-service). These can also be added using the same configuration mechanism.
```

Always better to use the global link and let it redirect as needed.

docs/other-models.md
```diff
@@ -101,6 +101,33 @@ And confirm they were logged correctly with:
...
+
+The `model_id` is the name LLM will use for the model. The `model_name` is the deployment name of the model on Azure OpenAI Service. 
+
+The `api_base` will be the Azure OpenAI Service endpoint. The `api_key_name` is the name of the Azure OpenAI Service API key stored using the {ref}`api-keys` feature.
+
+You must add `api_version` to use a specific version of the Azure OpenAI Service API.
+
+Set `api_type` to `azure` to specify that the OpenAI client library should use the Azure OpenAI Service API.
```
pamelafox, line 114: Worth noting that "azure_ad" is not supported, as far as I can tell (keyless auth with AAD).

pamelafox:
```suggestion
Set `api_type` to `azure` to specify that the OpenAI client library should use the Azure OpenAI Service API. Note that the type of `azure_ad` (keyless auth) is not yet supported.
```

docs/other-models.md
```diff
@@ -101,6 +101,33 @@ And confirm they were logged correctly with:
...
+
+You must also add `api_engine` to specify the base model to use, eg: `gpt-35-turbo-16k`
+
+For example, to add the `mygpt35` model (a `gpt-35-turbo-16k` model) deployed on Azure OpenAI Service, using the `2023-05-15` version of the API, with the account endpoint `https://my-openai-account.openai.azure.com/`, add this to your `extra-openai-models.yaml` file:
+
+
+```yaml
+- model_id: mygpt
+  model_name: mygpt35
```
pamelafox, line 123:
```suggestion
  model_name: gpt-35-turbo
```

This seems to have model and deployment reversed.

docs/other-models.md
```diff
@@ -101,6 +101,33 @@ And confirm they were logged correctly with:
...
+  api_base: https://my-openai-account.openai.azure.com/
+  api_key_name: azure
+  api_version: 2023-05-15
+  api_type: azure
+  api_engine: gpt-35-turbo-16k
```
pamelafox, line 128:
```suggestion
  api_engine: mygptchat
```

docs/other-models.md
```diff
@@ -101,6 +101,33 @@ And confirm they were logged correctly with:
...
+
+The `model_id` is the name LLM will use for the model. The `model_name` is the deployment name of the model on Azure OpenAI Service. 
+
+The `api_base` will be the Azure OpenAI Service endpoint. The `api_key_name` is the name of the Azure OpenAI Service API key stored using the {ref}`api-keys` feature.
+
+You must add `api_version` to use a specific version of the Azure OpenAI Service API.
+
+Set `api_type` to `azure` to specify that the OpenAI client library should use the Azure OpenAI Service API.
+
+You must also add `api_engine` to specify the base model to use, eg: `gpt-35-turbo-16k`
```
pamelafox, line 116:
```suggestion
You must also add `api_engine` to specify the name of the deployment.
```

docs/other-models.md
```diff
@@ -101,6 +101,33 @@ And confirm they were logged correctly with:
 llm logs -n 1
 ```
 
+### Azure OpenAI Service
+
+LLM supports OpenAI models deployed via [Azure OpenAI Service](https://azure.microsoft.com/en-us/products/ai-services/openai-service). These can also be added using the same configuration mechanism.
+
+The `model_id` is the name LLM will use for the model. The `model_name` is the deployment name of the model on Azure OpenAI Service.
```
pamelafox, line 108:
```suggestion
The `model_id` is the name LLM will use for the model. The `model_name` is the name of the model on Azure OpenAI Service, like `gpt-35-turbo-16k`.
```

pamelafox:
This now works for me after incorporating my suggestions.

pamelafox (changes requested):
See suggestions

sanzgiri:
@pamelafox - how are you specifying the azure openai key? Is it through the same env variable "OPENAI_API_KEY"? And both api_type and api_key_name are set to "azure"?

pamelafox:
I set the key using the key feature here:
https://llm.datasette.io/en/stable/setup.html#api-keys

Since I used the key name of "azure", then yes, my api_key_name was "azure", but in theory you might set it with a different name.

Here's what my configuration looks like in ~/Library/Application Support/io.datasette.llm/extra-openai-models.yaml :

- model_id: azure-gpt-35
  model_name: gpt-35-turbo
  api_base: https://cog-24ccg67vinxu2.openai.azure.com
  api_key_name: azure
  api_version: 2023-05-15
  api_type: azure
  api_engine: chat

sanzgiri:
@pamelafox - thanks! I was not setting up the key, works for me now.

DaniloNC:
I haven't been able to use Azure OpenAI with the instructions provided here.
See https://github.com/simonw/llm/issues/433 and  https://github.com/simonw/llm/issues/413

Since `openai` python package > 1.0.0, the `AzureOpenAI` class is required.

There is a pending PR here:
https://github.com/simonw/llm/pull/414

PR #338, opened 2023-11-09
Cli short help arg (-h)

nikitajz:
It is more convenient to use the `help` short counterpart `-h` for learning the `llm` tool interactively.

I've updated the docs as well (including autogenerated via `just cog`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,10 +107,10 @@ cat mycode.py | llm -s "Explain this code"
 
 ## Help
 
-For help, run:
+For help use `-h/--help` option:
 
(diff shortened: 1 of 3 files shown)


nikitajz:
Here is a workaround (hack) if someone looking for a solution.
Add the following alias to your shell config, e.g. `.zshrc`:
```
alias llm='function _llm(){ command llm $(echo "$@" | sed "s/-h/--help/g"); }; _llm'
```
It simply replaces `-h` to `--help`

Don't forget to reload the config or start a new shell

PR #357, opened 2023-11-27
minor enhancement proposal: Allows specifying custom conversation id

roblg:
When integrating into multi-step/unattended scripts, specifying a custom conversation ID in the initial call makes it easier to chain conversations. This avoids the need to look up the last autogenerated conversation ID for a given prompt/model combo, since it's known in advance.

I've done basic testing of:

* new conversation
* new conversation w/ custom ID
* resume conversation w/ autogen and custom ID

(I know this is out of the blue. No urgency, and no hard feelings if you disagree with the idea or the approach. I just wanted to write some code today.)

### DIFF
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -118,7 +118,8 @@ def cli():
     "conversation_id",
     "--cid",
     "--conversation",
-    help="Continue the conversation with the given ID.",
+    help="Continue the conversation with the given ID. "
+         "If no conversation with that ID exists, a new one will be created.",
 )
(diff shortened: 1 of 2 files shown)


neomanic:
I like this concept, because I've shared my llm chats enough with regular folks that I started automating the process to convert them into html and/or pdf, with `llm logs list -c > conversation.md` as the first step.

And a slug / summary as the conversation ID would present nicely, and reduce cleaning up the Markdown output.

Issue #315, opened 2023-10-26
Homebrew release can't install plugins that use PyTorch (no PyTorch for Python 3.12 yet)


jernst:
macOS Sonoma 14.0, M1 Pro

```
$ brew install llm
<skipped, no issues>
$ llm
Traceback (most recent call last):
  File "/opt/homebrew/bin/llm", line 5, in <module>
    from llm.cli import cli
  File "/opt/homebrew/Cellar/llm/0.11_1/libexec/lib/python3.12/site-packages/llm/__init__.py", line 18, in <module>
    from .plugins import pm
  File "/opt/homebrew/Cellar/llm/0.11_1/libexec/lib/python3.12/site-packages/llm/plugins.py", line 37, in <module>
    mod = importlib.import_module(plugin)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.0/Frameworks/Python.framework/Versions/3.12/lib/python3.12/importlib/__init__.py", line 90, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/llm/0.11_1/libexec/lib/python3.12/site-packages/llm/default_plugins/openai_models.py", line 6, in <module>
    import openai
  File "/opt/homebrew/Cellar/llm/0.11_1/libexec/lib/python3.12/site-packages/openai/__init__.py", line 19, in <module>
    from openai.api_resources import (
  File "/opt/homebrew/Cellar/llm/0.11_1/libexec/lib/python3.12/site-packages/openai/api_resources/__init__.py", line 1, in <module>
    from openai.api_resources.audio import Audio  # noqa: F401
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/llm/0.11_1/libexec/lib/python3.12/site-packages/openai/api_resources/audio.py", line 4, in <module>
    from openai import api_requestor, util
  File "/opt/homebrew/Cellar/llm/0.11_1/libexec/lib/python3.12/site-packages/openai/api_requestor.py", line 24, in <module>
    import requests
[...]

PR #370, opened 2023-12-02
Add homebrew install instruction

cander:
There are instructions about how to upgrade when using Homebrew, but not how to install. (very trivial)

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -10,6 +10,10 @@ Or using [pipx](https://pypa.github.io/pipx/):
 ```bash
 pipx install llm
 ```
+Or if you use [Homebrew](https://brew.sh/) for macOS:
+```bash
+brew install llm
+```
 
 ## Upgrading to the latest version
 


simonw:
I stopped recommending Homebrew deliberately a couple of months ago:
- https://github.com/simonw/llm/issues/315#issuecomment-1783697556

Reason is that the Homebrew build currently uses Python 3.12, which means that plugins that use PyTorch (such as `llm-sentence-transformers`) can't be installed using that Homebrew package.

This sucks. Hopefully PyTorch will release a Python 3.12 build soon, at which point it stops being a problem:
- https://github.com/pytorch/pytorch/issues/110436

simonw:
Hmm... that said, it looks like we do have a workaround:
- https://github.com/simonw/llm/issues/315#issuecomment-1879741434

So I could go back to recommending Homebrew but clearly document that workaround for when people need it.

simonw:
That workaround works - running it in CI over here now to prove it:
- https://github.com/simonw/llm-sentence-transformers/pull/13

cander:
Mostly, I was hoping for consistency - either install and update Hombrew, or no mention of Homebrew. Personally, I'm OK either way.

simonw:
I'm going to handle this in this issue instead:
- https://github.com/simonw/llm/issues/397

PR #395, opened 2024-01-25
Add ollama to plugin directory

Till--H:
Link to plugin that integrates with Ollama


simonw:
Thanks, neat plugin!

simonw:
Now live here: https://llm.datasette.io/en/stable/plugins/directory.html

Issue #378, opened 2023-12-17
Fix deprecation warning 'pkg_resources is deprecated as an API'


rupurt:
```
/Users/alex/workspace/rupurt/llm-http-api/.venv/lib/python3.11/site-packages/llm/plugins.py:3: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
    import pkg_resources

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

simonw:
Datasette has the same warning. About time I figured this one out.

Started by getting some clues from ChatGPT: https://chat.openai.com/share/44f23178-9ba0-4f32-bbc3-ea16d39ea7ff

After some hacking got to this point:
```
$ LLM_LOAD_PLUGINS=llm-cluster llm plugins                       
[
  {
    "name": "llm-cluster",
    "hooks": [
      "register_commands"
    ],
    "version": "0.2"
  }
]
$ LLM_LOAD_PLUGINS=llm-cluster,llm-clip llm plugins
[
  {
    "name": "llm-cluster",
    "hooks": [
      "register_commands"
    ],
    "version": "0.2"
  },
  {
    "name": "llm-clip",
    "hooks": [
      "register_commands",
      "register_embedding_models"
    ],
    "version": "0.1"
  }
]
```

simonw:
Note that `importlib.metadata` - which replaces `pkg_resources` - is new in Python 3.8 https://docs.python.org/3/library/importlib.metadata.html

simonw:
I need to run tests that exercise `LLM_LOAD_PLUGINS`. It looks like I have a local test for that which I never checked in, in a file called `tests/test-llm-load-plugins.sh`:

```bash
#!/bin/bash
# This should only run in environemnts where both
# llm-cluster and llm-sentence-transformers are installed

PLUGINS=$(llm plugins)
echo "$PLUGINS" | jq 'any(.[]; .name == "llm-sentence-transformers")' | \
  grep -q true || ( \
    echo "Test failed: llm-sentence-transformers not found" && \
    exit 1 \
  )
# With the LLM_LOAD_PLUGINS we should not see that
PLUGINS2=$(LLM_LOAD_PLUGINS=llm-cluster llm plugins)
echo "$PLUGINS2" | jq 'any(.[]; .name == "llm-sentence-transformers")' | \
  grep -q false || ( \
[...]

PR #399, opened 2024-01-26
Run test-llm-load-plugins.sh

simonw:
Refs:
- #378


simonw:
https://github.com/simonw/llm/blob/75db4a53931017457d0915ffaf52f85b189b5ba5/.github/workflows/test.yml#L44-L47

OK, `llm-sentence-transformers` is a bad one to use here because installing it pulls in CUDA which is huge and takes ages.

Issue #409, opened 2024-01-27
Fix skipped tests on Windows


simonw:
> I'd like to fix those test failures, but I also want to ship a 0.13.1 release quickly. So I'm going to mark a bunch of tests as expected to fail on Windows.

_Originally posted by @simonw in https://github.com/simonw/llm/issues/408#issuecomment-1912855140_

simonw:
A bunch of tests are marked with this because they don't yet pass on Windows - as far as I can tell this is all for `/` v.s. `\` reasons, or possible`\r\n` v.s. `\n`.

Some details here:
- https://github.com/simonw/llm/pull/408#issuecomment-1912853200

The skip mark is:
https://github.com/simonw/llm/blob/8021e12aaafdb452c9efd7b8571690f94efd579b/tests/test_keys.py#L9

simonw:
https://github.com/simonw/llm/actions/runs/7674759409/job/20919925464 is a run on Windows that failed.

```
2024-01-27T00:05:25.4573685Z ============================= test session starts =============================
2024-01-27T00:05:25.4575339Z platform win32 -- Python 3.10.11, pytest-7.4.4, pluggy-1.4.0
2024-01-27T00:05:25.4576305Z rootdir: D:\a\llm\llm
2024-01-27T00:05:25.4577287Z configfile: pytest.ini
2024-01-27T00:05:25.4577983Z plugins: anyio-4.2.0, httpx-0.28.0
2024-01-27T00:05:25.4578470Z collected 177 items
2024-01-27T00:05:25.4578702Z 
2024-01-27T00:05:25.5591928Z tests\test_aliases.py .............                                      [  7%]
2024-01-27T00:05:25.7946034Z tests\test_chat.py .FFFFFF                                               [ 11%]
2024-01-27T00:05:25.8234872Z tests\test_cli_openai_models.py ..                                       [ 12%]
2024-01-27T00:05:26.1278927Z tests\test_embed.py ..............                                       [ 20%]
2024-01-27T00:05:39.7648455Z tests\test_embed_cli.py ................................................ [ 47%]
2024-01-27T00:05:45.1569619Z .F.............                                                          [ 55%]
2024-01-27T00:05:45.1634150Z tests\test_encode_decode.py ..                                           [ 57%]
2024-01-27T00:05:46.0982847Z tests\test_keys.py .FF...                                                [ 60%]
2024-01-27T00:06:06.8133367Z tests\test_llm.py .....FF................................                [ 82%]
[...]

PR #410, opened 2024-01-27
works on windows

matthewdeanmartin:
os.sep + path + skipping readline for windows.

Why I suggest skipping readline:

Me & Chat GPT (back when it thought its name was Assistant) made a pacman game in the terminal that ran on windows. It took win32 API calls. Textual does it better: https://github.com/Textualize/textual/blob/main/src/textual/drivers/win32.py

Also I just wasn't sure what the motivation was. When I run `llm` in the input field I can use the arrows just fine. The regular `input()` function allows for left/right arrows. Prefilling an input and up/down arrows would make sense.

Addresses issue : https://github.com/simonw/llm/issues/409

### DIFF
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -1,3 +1,5 @@
+import os
+
 import click
 from click_default_group import DefaultGroup
 from dataclasses import asdict
@@ -342,8 +344,11 @@ def chat(
     Hold an ongoing chat with a model.
     """
     # Left and right arrow keys to move cursor:
-    readline.parse_and_bind("\\e[D: backward-char")
-    readline.parse_and_bind("\\e[C: forward-char")
+    if os.name != "nt":
+        # I'm pretty sure this can't be done without win32 API calls.
+        # pyreadline is unsupported and no longer python 3.12 compatible.
+        readline.parse_and_bind("\\e[D: backward-char")
+        readline.parse_and_bind("\\e[C: forward-char")
     log_path = logs_db_path()
     (log_path.parent).mkdir(parents=True, exist_ok=True)
     db = sqlite_utils.Database(log_path)
--- a/tests/test_chat.py
+++ b/tests/test_chat.py
@@ -2,7 +2,6 @@ from click.testing import CliRunner
 import llm.cli
 from unittest.mock import ANY
 import pytest
-import sys
 
 
 def test_mock_model(mock_model):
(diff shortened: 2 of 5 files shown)


simonw:
Would `pyreadline3` work on Windows here though?
https://github.com/simonw/llm/blob/8021e12aaafdb452c9efd7b8571690f94efd579b/setup.py#L50

simonw:
Pushed a tiny fix for this failing test:
```
        if os.name != "nt":
            # Should be chmod 600
>           assert oct(keys_path.stat().st_mode)[-3:] == "600"

tests/test_keys.py:36: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <_pytest._py.path.Stat object at 0x7f598d774490>, name = 'st_mode'

    def __getattr__(self, name: str) -> Any:
>       return getattr(self._osstatresult, "st_" + name)
E       AttributeError: 'os.stat_result' object has no attribute 'st_st_mode'
```

matthewdeanmartin:
Oh that bot! It changed `mode` to `st_mode` and I didn't see it fail.

`pyreadline3` no longer throws on import, but doesn't handle that key. I get the impression that it has only reimplemented some of what readline can do.

```error
File "E:\github\llm\llm\cli.py", line 347, in chat
    readline.parse_and_bind("\\e[D: backward-char")
  File "E:\.virtualenv\llm-Ou002Y_x\Lib\site-packages\pyreadline3\rlmain.py", line 112, in parse_and_bind
    self.mode._bind_key(key, func)
  File "E:\.virtualenv\llm-Ou002Y_x\Lib\site-packages\pyreadline3\modes\basemode.py", line 181, in _bind_key
    keyinfo = make_KeyPress_from_keydescr(key.lower()).tuple()
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "E:\.virtualenv\llm-Ou002Y_x\Lib\site-packages\pyreadline3\keysyms\common.py", line 138, in make_KeyPress_from_keydescr
    raise IndexError("Not a valid key: '%s'" % keydescr)
IndexError: Not a valid key: '\e[d'
```

I still don't know where the left/right arrows are failing to work so I'm not sure about how to go about implementing it for windows.

PR #429, opened 2024-02-28
Fix support for Python template syntax

jeeger:
All dollar signs in a template's `prompt` or `system` values are interpreted as introducing `placeholders`. The documentation for `string.Template` says that doubling a dollar sign should escape it.

Due to the custom implementation of `Template.extract_vars`, this did not work, as it extracted a double dollar sign as an identifier with an empty named group. By using `string.Template.get_identifiers()` instead, this now works. Additionally, this automatically adds support for parenthesized identifiers as well.

### DIFF
--- a/docs/templates.md
+++ b/docs/templates.md
@@ -92,6 +92,8 @@ prompt: >
 ```
 The `prompt: >` causes the following indented text to be treated as a single string, with newlines collapsed to spaces. Use `prompt: |` to preserve newlines.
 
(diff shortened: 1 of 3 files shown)


simonw:
Tests failed on Python 3.8:
```
llm/templates.py:54: in Template
    def extract_identifiers(cls, template: string.Template) -> list[str]:
E   TypeError: 'type' object is not subscriptable
```
I'll fix that now.

PR #438, opened 2024-03-14
Remove sqlite-migrate configuration from mypy.ini

simonw:
Refs:
- https://github.com/simonw/sqlite-migrate/issues/8


simonw:
The tests will fail on this until I've shipped a release of `sqlite-migrate` with the `py.typed` file in it.

Issue #308, opened 2023-10-20
setting options in templates


RodolfoCastanheira:
Is there a way to set options in a template? 

Use case: I wrote a template to output code to the command line, usually a very low temperature is used for that. It seems that different use cases of various templates could make use of different options for their model.

obra:
@RodolfoCastanheira to answer your specific question - there is not *currently* a way to to do that, but I agree that it'd be a very useful feature.

Quantisan:
closed by #830

PR #441, opened 2024-03-17
Add support for options in templates

ksangers:
I discovered that I couldn't set up options for templates. Until now I just made aliases to combine templates with options, but I'd rather have it stored with the template. I saw that someone else ran into this as well (#308). So here's my attempt on implementing it. Options can be passed like this per template:
```yaml
options:
  temperature: 0.2
  # more options key: value
# the rest of the template, e.g. system: ...
# ...
```

Is this something that could be implemented?

### DIFF
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -178,7 +178,9 @@ def prompt(
             if var:
                 disallowed_options.append(option)
         if disallowed_options:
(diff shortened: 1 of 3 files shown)


ksangers:
@simonw Is this something you'd consider implementing? If so, need anything changed in order to merge this?

Quantisan:
closed by #830

PR #455, opened 2024-04-10
Remove Homebrew warning

abitrolly:
PyTorch Python 3.12 binaries are available

https://pypi.org/project/torch/#files


abitrolly:
@simonw rebased Homebrew update.

PR #469, opened 2024-04-22
enable $ without template variables in system prompts

henrynine:
while working with a system prompt for latex generation, i found that `$` characters in a template system prompt that don't refer to a template variable will cause an error since `Template.extract_vars` will return a list containing `None` values. by ignoring `None` when checking for missing template vars, this error is avoided. using `safe_substitute` over `substitute` should be fine here because there is already a check for missing variables, and allows the use of `$` characters that don't refer to template vars in a system prompt without raising an error for an invalid placeholder.

### DIFF
--- a/docs/templates.md
+++ b/docs/templates.md
@@ -27,9 +27,9 @@ llm --system 'Summarize this text in the voice of $voice' \
   --model gpt-4 -p voice GlaDOS --save summarize
 ```
 
-To use `$` characters that don't refer to a template variable, escape them with another `$`:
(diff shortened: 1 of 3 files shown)


vgel:
I ran into this as well when trying to use `$$` escape syntax, which the existing code doesn't handle correctly:

```
>>> t = string.Template("$$n")
>>> list(t.pattern.finditer(t.template))
[<re.Match object; span=(0, 2), match='$$'>]
```

https://github.com/simonw/llm/blob/d654c9521235a737e59a4f1d77cf4682589123ec/llm/templates.py#L52-L57

PR #475, opened 2024-04-26
add `llm keys remove [name]` command

harabat:
These commits add the `remove` functionality to `llm keys`, add a test for that option, and update the relevant docs section.

### DIFF
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -525,6 +525,33 @@ def keys_set(name, value):
     current[name] = value
     path.write_text(json.dumps(current, indent=2) + "\n")
 
+@keys.command(name="remove")
+@click.argument("name")
+def keys_remove(name):
+    """
+    Remove a key from the keys.json file
+
+    Example usage:
+
+    \b
+        $ llm keys remove openai
+        Key 'openai' removed
+    """
+    path = user_dir() / "keys.json"
+    if not path.exists():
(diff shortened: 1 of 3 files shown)


llm/cli.py
```diff
@@ -525,6 +525,33 @@ def keys_set(name, value):
...
+    """
+    path = user_dir() / "keys.json"
+    if not path.exists():
+        click.echo("No keys found")
+        return
+    try:
+        current = json.loads(path.read_text())
+    except json.decoder.JSONDecodeError:
+        click.echo("Error reading keys file")
+        return
```
AlexanderYastrebov, line 548: I think try could be ommited like in `keys_list`.

llm/cli.py
```diff
@@ -525,6 +525,33 @@ def keys_set(name, value):
...
+    if name not in current:
+        click.echo(f"Key '{name}' not found")
+        return
```
AlexanderYastrebov, line 551: I think this should return error (useful for scripting).

llm/cli.py
```diff
@@ -525,6 +525,33 @@ def keys_set(name, value):
...
+    del current[name]
```
AlexanderYastrebov, line 552: This maybe should check for special note key `"// Note"` like keys_list does.

PR #496, opened 2024-05-16
Update openai-models.md to include 'model_name' for extra model config

sagarjauhari:
Errors out without `model_name`


tavisrudd:
Same issue and an updated fix: https://github.com/simonw/llm/pull/547

PR #501, opened 2024-05-23
Update openai-models.md

bcardiff:
I think there are some rendering issues at https://llm.datasette.io/en/stable/openai-models.html

Screenshot 2024-05-23 at 13 05 47

From the GitHub preview it seems this change should be enough, but I am not sure how cog will interact with it.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,6 +1,6 @@
 # LLM
 
-![GitHub Repo stars](https://img.shields.io/github/stars/simonw/llm)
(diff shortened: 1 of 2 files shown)


bcardiff:
Maybe this is also needed so the cog generates the new line.

```patch
--- cog.out("```\n{}```".format("\n".join(models)))
+++ cog.out("```\n{}\n```".format("\n".join(models)))
```

PR #517, opened 2024-06-19
Add `llm-deepseek` plugin

abrasumente233:
Based on the llm-fireworks plugin.


robtaylor:
@simonw any chance of this? would be great to use deepseek v3 from `llm`

abrasumente233:
> @simonw any chance of this? would be great to use deepseek v3 from `llm`

You can use it from OpenRouter, Fireworks or Together for now.

PR #522, opened 2024-06-26
Add llm-azure-api-embed

orangewise:
Hey Simon,

I've create a new embedding plugin for Azure OpenAI: https://github.com/orangewise/llm-azure-api-embed/.

Maybe you can add it to the list?

Thanks!
Ronald


fabge:
hi @orangewise, i have written a plugin, which uses the AzureOpenAI class and allows using the llm cli with azure api keys (including the embedding api)
let me know if it works for you!
https://github.com/fabge/llm-azure/

PR #525, opened 2024-06-28
Update outdated reference to gpt-4-turbo

simondonohue:
Looks like this alias was overlooked in 8171c9a. This commit makes it
match with the usage of gpt-4o in the associated example.


simonw:
Thanks for this.

PR #545, opened 2024-08-05
add azure-llm plugin to allow using the Azure OpenAI API

fabge:
since `openai >= 1.*` using azure as backend means having to use the `AzureOpenAI` class of the [`openai` python package](https://pypi.org/project/openai/), instead of just using the `OpenAI` class. this plugin does just that.
https://github.com/fabge/llm-azure


fabge:
the llm-azure package is a nice addition as it enables using all models provided by Azure.

Issue #495, opened 2024-05-15
llm 0.14: Can't run <<llm chat>> on Windows 11


rsbohn:
**Description:**
After upgrading from version 0.13, the following lines of code fail to execute on a Windows 11 environment:

```python
readline.parse_and_bind("\\e[D: backward-char")
readline.parse_and_bind("\\e[C: forward-char")
```

This issue appears to be specific to Windows. On non-Windows platforms, these lines work as expected.

**Steps to Reproduce:**
1. Upgrade from version 0.13 to the latest version on a Windows 11 machine.
2. Execute code containing the lines:
    ```python
    readline.parse_and_bind("\\e[D: backward-char")
    readline.parse_and_bind("\\e[C: forward-char")
    ```

**Observed Behavior:**
The code fails on Windows 11, leading to errors or unexpected behavior.

    IndexError: Not a valid key: '\e[d'

**Workaround Implemented:**
I have temporarily resolved the issue using a platform check to ensure this code doesn't execute on Windows. Here is the workaround:

```python
import os
import readline

if os.name != 'nt':
    readline.parse_and_bind("\\e[D: backward-char")
    readline.parse_and_bind("\\e[C: forward-char")
```

**Relevant Code:**
The relevant section of the code can be found at this [link to cli.py, line 345](https://github.com/simonw/llm/blob/68df9721def4820e1b54311a032c19bac9e0e8ea/llm/cli.py#L345).

**Environment:**
- OS: Windows 11
- Version affected: Latest version after 0.13

**Expected Behavior:**
The code should work across all platforms, including Windows 11, without requiring modifications.

rsbohn:
See also https://github.com/simonw/llm/commit/1a4853d80e9543ce93bba1b8cdf45d7201cd7a05

designcomputer:
Thanks. This is a good temporary fix for the issue with 0.14 on Windows.

rsbohn:
Still a problem in 0.15.

Issue #376, opened 2023-12-14
llm chat - readline problems still present


kevinburke:
When I open `llm chat`, I expect that using the left and right arrow keys will navigate the cursor but instead I get nasty ANSI codes printed to the screen. 

```
$ llm chat
Chatting with gpt-4
Type 'exit' or 'quit' to exit
Type '!multi' to enter multiple lines, then '!end' to finish
> foo bar ^[[D^[[D^[[D^[[D^[[D
```

I am using the latest version of llm, commit 5cc8efe98f49b3aaeddf0c05b7827efda411a916 (the readline support commit), installed by checking out this commit locally, starting a virtualenv and running "python setup.py install", then running e.g. `./venv/bin/llm`. 

I also verified that my Python version has readline support:

```
$ python
Python 3.10.13 (main, Nov  7 2023, 20:58:26) [Clang 15.0.0 (clang-1500.0.40.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import readline
>>> print(readline.__doc__)
Importing this module enables command line editing using GNU readline.
```

Let me know what else I can do to help troubleshoot.

kugurerdem:
I had the same experience. Running ``rlwrap llm chat`` resolved the issue for me when I wanted to chat through llm. However, it's a bit unfortunate that this great program has issues with readline.

Personally, I liked this program a lot and would like to see this problem addressed and solved.

simonw:
Working with ChatGPT: https://chat.openai.com/share/b5e1a34e-e365-4455-8d17-265ef80101c0 I found this magic incantation which seems to do the right thing:

```python
    readline.parse_and_bind("\\e[D: backward-char")
    readline.parse_and_bind("\\e[C: forward-char")
```
Weirdly obscure!

simonw:
Demo of that fix:

![readline](https://github.com/simonw/llm/assets/9599/65c790e4-397a-4dab-8610-2ccaa1559b28)

Issue #355, opened 2023-11-23
Support arrow cursors in "llm chat" mode


simonw:
https://twitter.com/josh_m/status/1727472264688677314

>  llm doesn't like arrow keys in a (I think) stock macOS terminal, it just inserts control chars.

simonw:
This seems to fix it:
```diff
diff --git a/llm/cli.py b/llm/cli.py
index af37feb..3fa2ecc 100644
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -28,6 +28,7 @@ from .plugins import pm
 import base64
 import pathlib
 import pydantic
+import readline
 from runpy import run_module
 import shutil
 import sqlite_utils
@@ -340,6 +341,7 @@ def chat(
     """
     Hold an ongoing chat with a model.
     """
+    readline.parse_and_bind("")
     log_path = logs_db_path()
     (log_path.parent).mkdir(parents=True, exist_ok=True)
     db = sqlite_utils.Database(log_path)
```

simonw:
![readline](https://github.com/simonw/llm/assets/9599/0c760955-f0d9-4805-a6b9-9d60a701316f)

simonw:
This was an incomplete fix, but I addressed it properly here:
- #376

PR #560, opened 2024-08-27
readline bindings only on mac and windows. fixes simonw/llm#516 simonw/llm#582 

### LINKED ISSUE #516: backslash cannot be used in `llm chat`
Hopefully this is not something stupid on my part, but I cannot seem to be able to use the backslash `\` in `llm chat` (not in single nor in multiline mode).
Both typing interactively and pasting text have this problem; in the latter case the backslash is simply taken out.

As you can imagine this is problematic when pasting code.

The workaround is to use a literal `Ctrl-v` before `\` then it appears, but when the prompt is long, with snippets of code, I have to open a scratch buffer in the editor and prefix each backslash with a literal `Ctrl-v` char which is error prone and something that in principle should not be asked to the user (in neovim you can type `ctrl-v` twice in insert mode).

`llm prompt` does not seem to have this problem.

grota:
fixes #516

### DIFF
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -860,12 +860,12 @@ def chat(
     Hold an ongoing chat with a model.
     """
     # Left and right arrow keys to move cursor:
-    if sys.platform != "win32":
-        readline.parse_and_bind("\\e[D: backward-char")
-        readline.parse_and_bind("\\e[C: forward-char")
-    else:
+    if sys.platform == "win32":
         readline.parse_and_bind("bind -x '\\e[D: backward-char'")
         readline.parse_and_bind("bind -x '\\e[C: forward-char'")
+    elif sys.platform == "darwin":
+        readline.parse_and_bind("\\e[D: backward-char")
+        readline.parse_and_bind("\\e[C: forward-char")
     log_path = pathlib.Path(database) if database else logs_db_path()
     (log_path.parent).mkdir(parents=True, exist_ok=True)
     db = sqlite_utils.Database(log_path)


grota:
@simonw I modified the PR to import readline and bind only on Darwin, this fixes the following 4 open issues: simonw/llm#516 simonw/llm#582 simonw/llm#585 simonw/llm#495 

As commented elsewhere, digging through the history it seems like readline is used to fix this issue #376 which is the proper fix for #355 (so basically, the standard terminal on osx).

It's a simple PR, can you please give it a look?

knorthover:
I just hit this issue with readline on Windows 10.  Claude Sonnet suggested this patch to use the codes readline recognizes for Windows.  It works.

``` 
    import platform
    if platform.system() == 'Windows':
        readline.parse_and_bind("Left: backward-char")
        readline.parse_and_bind("Right: forward-char")
    else:
        readline.parse_and_bind("\\e[D: backward-char")
        readline.parse_and_bind("\\e[C: forward-char") ```

grota:
Based on the analysis above, the two bindings were introduced to address an issue on macOS (darwin), specifically with the default iTerm app.

Considering the existing issues, I believe it’s fair to say that on Linux and Windows, these bindings have caused more problems than they've solved. So I stand by my original PR which removes the import and bindings on those platforms.

In fact, I don't think hardcoding these bindings is a good approach even on darwin. It wouldn’t surprise me if similar issues arose on macOS with other terminal emulators.

@simonw I'm asking you once again if can you please have a look at this simple PR of mine

grota:
@simonw just FYI, I rebased and rewritten the PR to perform readline bindings only on darwin and windows, as per the new code by @sukhbinder in #646.

grota:
closing, this issue is solved with a recent version of click

PR #572, opened 2024-09-12
Update openai-models.md to fix 'Adding more OpenAI models' section

### LINKED ISSUE #333: Error if
The [documentation for adding additional OpenAI models](https://llm.datasette.io/en/stable/other-models.html#adding-more-openai-models) gives this example of adding a model by its API ID to extra-openai-models.yaml:
```
- model_id: gpt-3.5-turbo-0613
  aliases: ["0613"]
```
But if you try to use that llm will get an error, because there it has no `model_name`.

The real problem is that line 50 of openai_models.py is `model_name = extra_model["model_name"]` which tries to directly access the "model_name" key, instead of `model_name = extra_model.get("model_name")` which safely handles failure.

Quantisan:
fixes #333

### DIFF
--- a/docs/openai-models.md
+++ b/docs/openai-models.md
@@ -106,9 +106,11 @@ Let's say OpenAI have just released the `gpt-3.5-turbo-0613` model and you want 
 
 ```yaml
 - model_id: gpt-3.5-turbo-0613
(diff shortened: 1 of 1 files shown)


simonw:
Thanks!

PR #580, opened 2024-10-08
Fix broken tests.  Drop python 3.8.

rectalogic:
A number of tests are failing on main, apparently because tests always use the latest version of every dependency (no `Pipfile.lock`) and e.g. `pytest_httpx` has breaking changes.

Also `pytest_httpx` now requires python>=3.9 so I dropped 3.8 support.

What do you think about switching from `setup.py` to isolated builds using `uv` and checking in a `uv.lock` so tests don't fail over time?

```
=========================== short test summary info ============================
FAILED tests/test_keys.py::test_uses_correct_key - assert 1 == 0
FAILED tests/test_templates.py::test_template_basic[boo-extra_args3-None-None-Error: Cannot use -t/--template and --system together] - TypeError: HTTPXMock.reset() got an unexpected keyword argument 'assert_all...
FAILED tests/test_templates.py::test_template_basic[prompt: 'Say $hello'-extra_args4-None-None-Error: Missing variables: hello] - TypeError: HTTPXMock.reset() got an unexpected keyword argument 'assert_all...
ERROR tests/test_keys.py::test_uses_correct_key - AssertionError: The following requests were not expected:
ERROR tests/test_templates.py::test_template_basic[boo-extra_args3-None-None-Error: Cannot use -t/--template and --system together] - AssertionError: The following responses are mocked but not requested:
ERROR tests/test_templates.py::test_template_basic[prompt: 'Say $hello'-extra_args4-None-None-Error: Missing variables: hello] - AssertionError: The following responses are mocked but not requested:
============ 3 failed, 181 passed, 638 warnings, 3 errors in 3.23s =============
error: Recipe `test` failed on line 10 with exit code 1
```

PR #581, opened 2024-10-10
Support tool calling.

rectalogic:
This supports LLM tool calling. Plugins can register new tools (python callables that are introspected to generate JSON schema), and models are provided access to installed tools. Adds tool calling to the openai Chat model.

Bundles a default `read_files` tool.
Here is a sample `brave_search` tool in a plugin
https://github.com/rectalogic/llm-tools/blob/dev/src/llm_tools/brave_search.py

* add register_tools hookspec to allow plugins to register tools
* add @llm.Tool decorator that introspects a callables type signature and generates a JSON schema
* expose installed tools to Models via new `tools` Model property
* implement tool calling in the openai Model
* register a default `read_files` tool

Note: this pins pytest_httpx until broken tests are fixed in
  https://github.com/simonw/llm/pull/580
Note: this requires openai>=1.40.0
Note: this drops 3.8 since a number of dependencies no longer
  support it, and adds 3.13 support


rectalogic:
Hmm, tool call responses can result in more tool calls, so I made openai loop.
e.g.
```sh-session
> llm chat --enable-tools
Chatting with gpt-4o-mini
Type 'exit' or 'quit' to exit
> Search Brave for the query "hurricane milton" and write the first line of the esponse to /tmp/hurricane
Tool: brave_search(query=hurricane milton)
Tool: bash(command_line=echo 'Milton made landfall near Siesta Key, Florida, as a dangerous Category 3 hurricane before weakening as it cut through the state.' > /tmp/hurricane)
result
The first line of the search result for "hurricane milton" has been successfully written to `/tmp/hurricane`. The content states:

"Milton made landfall near Siesta Key, Florida, as a dangerous Category 3 hurricane before weakening as it cut through the state."
```

simonw:
I'm working on this feature over here now: #898

simonw:
There are some great ideas in here, sorry for not providing a full review!

PR #594, opened 2024-10-29
Example of attachment with type

simonw:
Need to test this before merging it.

### DIFF
--- a/docs/python-api.md
+++ b/docs/python-api.md
@@ -54,7 +54,7 @@ response = model.prompt(
 
 ### Attachments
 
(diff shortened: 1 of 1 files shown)


docs/python-api.md
```diff
@@ -69,7 +69,18 @@ response = model.prompt(
     ]
 )
 ```
-Use `llm.Attachment(content=b"binary image content here")` to pass binary content directly.
+Use `llm.Attachment(content=b"binary image content here")` to pass binary content directly. You may optionqlly specify a content type - if tou do not it will be detected automatically:
```
erikdw, line 72:
```suggestion
Use `llm.Attachment(content=b"binary image content here")` to pass binary content directly. You may optionally specify a content type - if you do not it will be detected automatically:
```

PR #600, opened 2024-10-31
Allow passing of can_stream in openai_models.py

### LINKED ISSUE #599: Allow setting can_stream in extra-openai-models.yaml to allow for o1 over proxy
o1 support was added in response to 

- #570

However, this hardwires the streamability to named o1 models. I am accessing o1-preview via a (litellm) proxy, so I get a `'message': 'litellm.BadRequestError: AzureException BadRequestError - Error code: 400 - {\'error\': {\'message\': "Unsupported value: \'stream\' does not support true with this model. Only the default (false) value is supported.", \'type\': \'invalid_request_error\'`

I believe I need to be able to do this

```yaml
- model_name: openai/o1
  model_id: lbl/o1
  api_base: <MY PROXY>
  api_key_name: <MY KEY NAME>
  can_stream: false
```

however, `can_stream` is currently ignored

cmungall:
Fixes #599 

A longer term fix would be to use something like Pydantic so we don't repeat ourselves, but this would be a bit of a refactor

### DIFF
--- a/docs/other-models.md
+++ b/docs/other-models.md
@@ -45,6 +45,8 @@ You can set `api_key_name` to the name of a key stored using the {ref}`api-keys`
 
 Add `completion: true` if the model is a completion model that uses a `/completion` as opposed to a `/completion/chat` endpoint.
 
+If a model does not support streaming, add `can_stream: false` to disable the streaming option.
+
 Having configured the model like this, run `llm models` to check that it installed correctly. You can then run prompts against it like so:
 
 ```bash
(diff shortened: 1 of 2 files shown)


simonw:
This fix doesn't look quite right to me.

The code currently defaults to assuming `can_stream = True`:

https://github.com/simonw/llm/blob/1a60fa16674ee5d70a359018dbfd545f6689cbfe/llm/default_plugins/openai_models.py#L283-L293

With this change that will switch to assuming `False` (or `None`) for any models configured in the `extra-openai-models.yaml` file.

I think this should only take effect if `can_models: False` in the YAML.

Issue #507, opened 2024-06-10
Ability to execute prompts against an asyncio (non-blocking) API


simonw:
Datasette in particular needs a neat way to run different models via an `await ...` async method. Having a way for model plugins to provide this - especially the API based plugins like OpenAI and Claude - would be really helpful and would unblock a bunch of Datasette plugins that could benefit from an LLM abstraction layer.

simonw:
Two options:

1. Plugins have to specifically upgrade to offer this
2. Works for all plugins - those that have not yet upgraded (like most of the local model ones) can run in a thread and LLM will provide an `async` interface to them anyway

I'm going to build 1. I will treat 2. as an optional nice-to-have in the future.

I expect that any of the API-driven plugins which do not use a client library (like `llm-mistral`) will want a mechanism where they can define how the HTTP requests work and then some abstraction makes that work with both blocking and asyncio `httpx`.

simonw:
Two options for the API design:

1. Keep the current model classes, but add `await model.aprompt(...)` style methods, similar to the Django ORM's `await author.books.afirst()` style of API design
2. Offer a separate async class - so `model = llm.get_async_model("gpt-4o")` and then `async for token in model.prompt(...)`

I'm torn on these two. Not sure how to decide.

simonw:
I used Claude as an electric bicycle for the mind and I've decided to go with the separate async class. I think having a separate class will make things like type hints more obvious. https://gist.github.com/simonw/f69e6a1fe21df5007ce038dfe91c62f4

simonw:
I'm going to add a `llm prompt ... --async` option purely for my own testing which switches to the async version of a model.

simonw:
I'm going to have to do this for embedding models, too.

simonw:
This may be tricky: https://github.com/simonw/llm/blob/fe1e09706f69a4934430d2f2dadc354ffc4aa4dc/llm/cli.py#L361-L363

That's a reminder that sometimes we do `model.prompt()` and sometimes we do `conversation.prompt()` - which means we'll need an async mechanism for conversations, too - or some kind of thing where we can get an async alternative version of a conversation.

One option: instead of `conversation.prompt()` we go back to `model.prompt()` and add `conversation=` as a keyword argument.

Reminder though: the reason we have `conversation.prompt()` is the idea was to add tool support as a feature of conversations, since when you run a function you may end up having a single prompt that executes multiple rounds against the LLM as it executes functions and sends in their outputs.

[...]

Issue #610, opened 2024-11-06
Abstract out token usage numbers


simonw:
Most APIs return the number of input and output tokens used (and sometimes a more detailed breakdown of those categories). I want to provide an abstraction over those to make it easier to implement tools on top of LLM that do token accounting.

simonw:
The most complex form of token accounting right now is OpenAI - their usage blocks look like this as-of yesterday when they added [Predicted Outputs](https://simonwillison.net/2024/Nov/4/predicted-outputs/):
```json
{
  "completion_tokens": 9,
  "prompt_tokens": 8,
  "total_tokens": 17,
  "prompt_tokens_details": {
    "cached_tokens": 0,
    "audio_tokens": 0
  },
  "completion_tokens_details": {
    "reasoning_tokens": 0,
    "audio_tokens": 0,
    "accepted_prediction_tokens": 0,
    "rejected_prediction_tokens": 0
  }
}
```

simonw:
If my eventual goal is to support code that can calculate dollar spends I'll also need to consider how I store the fact that a prompt was executed in batch mode, which often provides a 50% discount. LLM doesn't have a mechanism for batch mode yet but it's definitely a potentially useful feature.

simonw:
Given the complexity of the OpenAI response I'm tempted to add a JSON column to store this. But the above example would actually be a waste of JSON, since the only things that actually matter in there are `completion_tokens=9` and `prompt_tokens=8` - everything else is safe to ignore.

I could have columns for `input_tokens` and `output_tokens` (better names than completion and prompt in my opinion) and a nullable JSON column for `token_details` which I would leave blank in this case but populate for cases when those other token categories are worth recording.

simonw:
Gemini 1.5 Pro usage looks like this (according to `llm logs -m gemini-1.5-pro-latest --json`):

```json
"usageMetadata": {
  "promptTokenCount": 14,
  "candidatesTokenCount": 26,
  "totalTokenCount": 40
}
```json
Anthropic is even simpler:
```
"usage": {
  "input_tokens": 8,
  "output_tokens": 18
}
```

**Comment by simonw** [comment]
`xAI/grok-beta`:

```json
"usage": {
  "completion_tokens": 349,
  "prompt_tokens": 12,
  "total_tokens": 361
}
```
`lambdalabs/hermes3-405b`:
```json
"usage": {
  "prompt_tokens": 15,
  "completion_tokens": 1,
  "total_tokens": 16,
[...]

PR #613, opened 2024-11-06
llm.get_async_model(), llm.AsyncModel base class and OpenAI async models

simonw:
Refs:
- #507

Still to figure out:

- [x] How do plugins register their async versions? Might need a `register_async_models()` hook.
- [x] Get another plugin working - Anthropic or Gemini or Ollama would be good
- [x] How does logging to the DB work? That's currently a method on `Model`, needs to move out and an async version needs to be considered.
- [x] How about loading FROM the database? Will that work for things like attachments too?
- [x] Make mypy happy
- [x] Refactor to avoid generics
- [x] Testing for all of this
- [x] Python API documentation
- [x] Documentation for writing async model plugins

----

Punted on these:

- Are there any other weird knock-on impacts of this I haven't considered yet?
- How should embeddings work?  https://github.com/simonw/llm/issues/628
- Maybe try getting a `llm-gguf` plugin to work with this? Might help confirm the API design further by showing it working with a blocking plugin (maybe via a run-in-thread hack). [SmolLM2](https://simonwillison.net/2024/Nov/2/smollm2/) might be good for this.

### DIFF
--- a/llm/default_plugins/openai_models.py
+++ b/llm/default_plugins/openai_models.py
@@ -526,7 +526,7 @@ class AsyncChat(_Shared, AsyncModel):
             )
             response.response_json = remove_dict_none_values(completion.model_dump())
             yield completion.choices[0].message.content
-        response._prompt_json = redact_data_urls({"messages": messages})
+        response._prompt_json = redact_data({"messages": messages})
 
 
 class Completion(Chat):
--- a/llm/default_plugins/openai_models.py
+++ b/llm/default_plugins/openai_models.py
@@ -270,14 +270,10 @@ class SharedOptions(llm.Options):
         return validated_logit_bias
 
 
-class Chat(Model):
+class _Shared:
     needs_key = "openai"
     key_env_var = "OPENAI_API_KEY"
-
     default_max_tokens = None
-
-    def get_async_model(self):
-        return AsyncChat(self.model_id, self.key)
 
     class Options(SharedOptions):
         json_object: Optional[bool] = Field(
@@ -370,40 +366,6 @@ class Chat(Model):
             messages.append({"role": "user", "content": attachment_message})
         return messages
 
-    def execute(self, prompt, stream, response, conversation=None):
-        if prompt.system and not self.allows_system_prompt:
-            raise NotImplementedError("Model does not support system prompts")
-        messages = self.build_messages(prompt, conversation)
-        kwargs = self.build_kwargs(prompt, stream)
-        client = self.get_client()
-        if stream:
-            completion = client.chat.completions.create(
-                model=self.model_name or self.model_id,
-                messages=messages,
-                stream=True,
-                **kwargs,
-            )
-            chunks = []
-            for chunk in completion:
-                chunks.append(chunk)
-                try:
-                    content = chunk.choices[0].delta.content
-                except IndexError:
-                    content = None
-                if content is not None:
-                    yield content
-            response.response_json = remove_dict_none_values(combine_chunks(chunks))
-        else:
-            completion = client.chat.completions.create(
-                model=self.model_name or self.model_id,
-                messages=messages,
-                stream=False,
-                **kwargs,
-            )
-            response.response_json = remove_dict_none_values(completion.model_dump())
-            yield completion.choices[0].message.content
-        response._prompt_json = redact_data_urls({"messages": messages})
-
     def get_client(self, async_=False):
         kwargs = {}
         if self.api_base:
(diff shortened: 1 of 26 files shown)


simonw:
Turns out Claude can help resolve merge conflicts: https://gist.github.com/simonw/84104386e788e2797a93581c7985ea32

simonw:
Nasty test failure:

```
FAILED tests/test_aliases.py::test_set_alias[gpt-3.5-turbo] - RecursionError: maximum recursion depth exceeded while calling a Python object
```

simonw:
Got the existing tests passing again!

simonw:
I broke this, found out while trying to write tests for it:

```pycon
>>> import asyncio
>>> import llm
llm.model = llm.get_async_model("gpt-4o-mini")-4o-mini")
>>> model
<AsyncChat 'gpt-4o-mini'>
>>> async for token in model.prompt("describe a good dog in french"):
...     print(token, end="", flush=True)
... 
Traceback (most recent call last):
  File "/Users/simon/.pyenv/versions/3.10.4/lib/python3.10/asyncio/__main__.py", line 58, in runcode
    return future.result()
  File "/Users/simon/.pyenv/versions/3.10.4/lib/python3.10/concurrent/futures/_base.py", line 446, in result
    return self.__get_result()
  File "/Users/simon/.pyenv/versions/3.10.4/lib/python3.10/concurrent/futures/_base.py", line 391, in __get_result
    raise self._exception
  File "<console>", line 1, in <module>
  File "/Users/simon/Dropbox/Development/llm/llm/models.py", line 376, in __aiter__
    async for chunk in await self.model.execute(
TypeError: object async_generator can't be used in 'await' expression
```

simonw:
Found the fix
```diff
diff --git a/llm/models.py b/llm/models.py
index 25a016b..1e6c165 100644
--- a/llm/models.py
+++ b/llm/models.py
@@ -373,7 +373,7 @@ class AsyncResponse(_BaseResponse["AsyncModel", Optional["AsyncConversation"]]):
                 yield chunk
             return
 
-        async for chunk in await self.model.execute(
+        async for chunk in self.model.execute(
             self.prompt,
             stream=self.stream,
             response=self,
```

simonw:
One last mypy problem to figure out:
```
  mypy
llm/models.py:390: error: "Coroutine[Any, Any, AsyncIterator[str]]" has no attribute "__aiter__" (not async iterable)  [attr-defined]
llm/models.py:390: note: Maybe you forgot to use "await"?
Found 1 error in 1 file (checked 15 source files)
error: Recipe `lint` failed on line 22 with exit code 1
```

simonw:
Trying this:
```bash
files-to-prompt llm/models.py -c | llm -m o1-preview 'llm/models.py:390: error: "Coroutine[Any, Any, AsyncIterator[str]]" has no attribute "__aiter__" (not async iterable)  [attr-defined]
llm/models.py:390: note: Maybe you forgot to use "await"?'
```

simonw:
The fix to all of my horrible `mypy` errors turned out to be doing this:
```python
class AsyncModel(_BaseModel["AsyncResponse", "AsyncConversation"]):
    def conversation(self) -> "AsyncConversation":
        return AsyncConversation(model=self)

    @abstractmethod
    async def execute(
        self,
        prompt: Prompt,
        stream: bool,
        response: "AsyncResponse",
        conversation: Optional["AsyncConversation"],
    ) -> AsyncGenerator[str, None]:
        """
        Returns an async generator that executes the prompt and yields chunks of text,
        or yields a single big chunk.
        """
        yield ""

```
Note the `yield ""` in the abstract method, which made it match that `AsyncGenerator[str, None]` signature.

`AsyncGenerator[str, None]` means it yields strings and does not have any special return when it finishes.

simonw:
I might make `Response.log_to_db()` a class method.

llm/models.py
```diff
@@ -7,13 +7,32 @@
...
 from abc import ABC, abstractmethod
 import json
 from pydantic import BaseModel
 from ulid import ULID
 
+ModelT = TypeVar("ModelT", bound=Union["Model", "AsyncModel"])
+ConversationT = TypeVar(
+    "ConversationT", bound=Optional[Union["Conversation", "AsyncConversation"]]
+)
+ResponseT = TypeVar("ResponseT")
```
simonw, line 33: I'm not happy about this at all, it's way too hard to understand.

llm/models.py
```diff
@@ -161,32 +220,56 @@ def __init__(
...
+                [row["id"]],
+            )
+        ]
+        return response
 
-    def _force(self):
-        if not self._done:
-            list(self)
 
+class Response(_BaseResponse["Model", Optional["Conversation"]]):
```
simonw, line 265: Also this - I'm going to try refactoring to not use generics like this.

llm/default_plugins/openai_models.py
```diff
@@ -406,38 +471,53 @@ def execute(self, prompt, stream, response, conversation=None):
...
+class AsyncChat(_Shared, AsyncModel):
+    needs_key = "openai"
+    key_env_var = "OPENAI_API_KEY"
+    default_max_tokens = None
+
+    class Options(SharedOptions):
+        json_object: Optional[bool] = Field(
+            description="Output a valid JSON object {...}. Prompt must mention JSON.",
+            default=None,
+        )
```
simonw, line 484: Can I refactor this to avoid duplication?

simonw:
... and after all of this, I'm thinking maybe a single model class with an `aexecute()` method might be better after all? I might try it just to see how it feels.

simonw:
I got this working:
```pycon
>>> await llm.get_async_model("gpt-4o-mini").prompt("hi")
'Hello! How can I assist you today?'
```
Using this:
```diff
diff --git a/llm/models.py b/llm/models.py
index 3ed61bf..8539d08 100644
--- a/llm/models.py
+++ b/llm/models.py
@@ -416,6 +416,9 @@ class AsyncResponse(_BaseResponse):
         await self._force()
         return self._start_utcnow.isoformat() if self._start_utcnow else ""
 
+    def __await__(self):
+        return self.text().__await__()
+
     @classmethod
     def fake(
         cls,

```

simonw:
> ```pycon
> >>> await llm.get_async_model("gpt-4o-mini").prompt("hi")
> 'Hello! How can I assist you today?'
> ```

This is an API design mistake. It means you can't get the `response` object, which is necessary for things like looking up usage tokens (when I implement #610).

simonw:
This is better:
```pycon
>>> import asyncio
>>> import llm
>>> m = llm.get_async_model("gpt-4o-mini")
>>> response = await m.prompt("say hi in spanish")
>>> response
<Response prompt='say hi in spanish' text='¡Hola!'>
>>> await response.text()
'¡Hola!'
```

PR #618, opened 2024-11-07
Keys in environment should precede static configuration

kfet:
This allows more specific methods to override more general ones. One can override persistent configurations (files) with temporary ones (env vars), and override both with one-time CLI parameters.

* CLI parameters are most explicit and specific to that execution
* Environment variables provide session-level configuration
* Config files offer system-wide/default settings

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@ venv
 *.egg-info
 .DS_Store
 .idea/
+.direnv/
+.envrc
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -8,3 +8,5 @@ venv
 *.egg-info
 .DS_Store
 .idea/
+.direnv/
+.envrc
```
kfet, line 12: just some more common ignores

PR #622, opened 2024-11-11
Update default model information to 4o-mini

tnorthcutt:
The default model is now 4o-mini; this change updates the usage page of the docs to reflect that


simonw:
Thanks! I missed those ones.

PR #634, opened 2024-11-16
Enhancement Proposal: Track Head (for backtracking)

FergusFettes:
Sometimes it would be nice to step back a single generation. To fix a typo or redo a generation. This PR makes that possible. This is done by adding a parent ID to the responses and adding a table to track the HEAD ID.

I can clean it up and add some tests, but first I wanted to check if this was something you would be amenable to.

Here is a demonstration of what it allows for:
```
> llm 'Hello World. Reply with "hello".'
hello
> llm --continue 'Magic word is pineapple. Confirm with "okay".'
okay
> llm --continue 'What is the magic word?'
pineapple
> llm head back
Head moved back to response 01jctw7f1a2xv2n8ckjyhxr3k7
> llm head back
Head moved back to response 01jctw6rh9pd2kvphs0sdcjz8e
> llm --continue 'What is the magic word?'
please
> llm head set 01jctw7f1a2xv2n8ckjyhxr3k7
Head is now at response 01jctw7f1a2xv2n8ckjyhxr3k7
> llm --continue 'What is the magic word?'
pineapple
```

I think this is useful on its own and not too big of a change. However, this would also make it possible to explore many branches of a conversation in parallel, in a [loom](https://github.com/socketteer/loom)-like manner. I would make that as an additional module, but for that to be possible the responses need parent ids, so at least that much would have to be added to the db I believe.

### DIFF
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -447,15 +447,23 @@ def chat(
         print("")
 
 
+def get_most_recent_active_conversation(db):
+    return next(db.query("""
+        SELECT conversation_id, MAX(datetime_utc) as last_active
+        FROM responses
+        GROUP BY conversation_id
+        ORDER BY last_active DESC
+        LIMIT 1
+    """), {}).get("conversation_id")
+
+
 def load_conversation(conversation_id: Optional[str]) -> Optional[Conversation]:
     db = sqlite_utils.Database(logs_db_path())
     migrate(db)
     if conversation_id is None:
         # Return the most recent conversation, or None if there are none
-        matches = list(db["conversations"].rows_where(order_by="id desc", limit=1))
-        if matches:
-            conversation_id = matches[0]["id"]
-        else:
+        conversation_id = get_most_recent_active_conversation(db)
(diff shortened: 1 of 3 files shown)


llm/cli.py
```diff
@@ -443,35 +444,185 @@ def chat(
...
+
+def get_response_parent(response, db):
+    if response.parent_id:
+        # Use explicit parent if set
+        return next(db.query("SELECT * FROM responses WHERE id = ?", [response.parent_id]), None)
+
+    if not response.conversation:
+        return {'id': None}
+
+    # Fall back to previous response in conversation
```
FergusFettes, line 477: get_response_parent will fall back to the default behaviour, so it will work like normal with existing dbs without having to populate the parent_id for existing responses.

FergusFettes:
I have started turning this into a plugin, but I have to modify the migrations and do a lot of monkey-patching so its probably not something most people should use. But if you like you can close this.

https://github.com/FergusFettes/llm-head

PR #637, opened 2024-11-18
Add `editor` command (and make it the default when no args provided)

oeo:
Adds an interactive editor mode to the LLM CLI:

I noticed `llm` was just hanging without any args and had a personal tool that allowed me to open my system default editor and sort of draft a more complex prompt and thought the feature was useful. 

- New `llm editor` command to compose prompts in a system editor
- Default behavior now opens an editor when no arguments are provided
- Supports specifying custom editor with `-e/--editor` flag

### Usage

```bash
# Open default editor
llm
llm editor

# Open with specific editor
llm editor -e vim
```

image

image

### DIFF
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -1,3 +1,4 @@
+import os
 import asyncio
 import click
 from click_default_group import DefaultGroup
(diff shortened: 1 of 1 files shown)


llm/cli.py
```diff
@@ -267,8 +312,8 @@ def read_prompt():
             and not attachments
             and not attachment_types
         ):
-            # Hang waiting for input to stdin (unless --save)
-            prompt = sys.stdin.read()
+            # Open editor instead of hanging
+            prompt = open_editor_for_prompt()
```
gdbing, line 316: I understand that this is what prompted the idea for this change in the first place but waiting for stdin is actually very useful, it allows piping other commands into llm e.g. `cat file.txt |llm -s "analyze the contents of this file"`

You should check if there is input waiting on stdin
e.g. try something like
```
if select.select([sys.stdin], [], [], 0.1)[0]: # Check stdin with 0.1 second timeout
    prompt = sys.stdin.read()
else:
    prompt = open_editor_for_prompt()
```

Issue #617, opened 2024-11-06
Fragments feature for passing references to text/URLs that get stored in a de-duplicated table


simonw:
The [attachments](https://github.com/simonw/llm/issues/587) feature makes it easy to attach video and audio from a URL or file while ensuring that even if that prompt component is used many times over it only ends up stored once in the database - in the `attachments` table using a `sha256` hash as the ID: https://github.com/simonw/llm/blob/febbc04fb6d847bd35ce5ba0156ef010f5894564/llm/models.py#L28-L39

I find myself wanting the same thing for blocks of text: I'd like to construct e.g. a context that contains the full docs for LLM and run multiple prompts against that without storing many copies of the same exact text in the database.

(I think this will also provide a useful hook for implementing Claude prompt caching.)

One possible solution would be to re-use the attachments mechanism, with a new default plain text attachment type that all models support:

```bash
llm -m gpt-4o-mini \
  -a https://gist.githubusercontent.com/simonw/f7b251a05b834a9b60edff7e06d31572/raw/3ad2fb03d57aad7b616ec40455c355ec51d2e3db/llm-docs.md \
  'how do I embed a file?'
```

----

Eventually decided on a `-f/--fragment` design:
- https://github.com/simonw/llm/issues/617#issuecomment-2480755035

simonw:
Pros of this approach:

- Reuses existing mechanism, no new tables or columns needed
- Neatly solves the need for reused long text prompts

Cons:

- A bit confusing: attachments were for multi-modal models, now we have a special type that does something else?
- Potentially confusing UI: `llm -a URL` could point to all kinds of text files, how do we know which ones get pasted into the prompt?

simonw:
Alternative approaches:
- Do nothing. Store long context prompts many many times in the same table.
- Introduce a new non-attachments mechanism for this, so e.g. the `responses` table can optionally use a foreign key to a new `reused_text` table which works a bit like attachments do right now.
- A mechanism where a `response` can point to some other response and say `I used the same prompt` as this one.

simonw:
Ideally whatever mechanism I use here should work for both `system` and regular prompts.

If using existing attachments we could invent artificial content types of `llm/prompt` and `llm/system` for this.

simonw:
Question: in my current database how many responses have duplicates in prompts or system prompts?

simonw:
Simple dupe detection query:
```bash
sqlite-utils "$(llm logs path)" "
[...]

PR #638, opened 2024-11-18
WIP: Fragments

simonw:
Refs:
- #617

TODO:

- [x] Make aliases such as `llm prompt -f alias` work
- [x] Document how to use `-f` option
- [x] `llm fragments list` needs non-JSON output (that truncates) - this should work similar to the new `llm logs --short` mode.
- [x] `llm fragments -q one -q two`
- [x] Make `-f fragment-hash` work in addition to `-f fragment-alias`
- [x] Documentation for the `llm fragments` family of commands
- [ ] Make sure `-f $URL` follows up to three redirects
- [ ] Fix the problem where `prompt_json` still duplicates the prompt - I can use https://pypi.org/project/condense-json/ for that
- [ ] Update `llm logs` to correctly display prompts using this new mechanism
  - [ ] Consider hiding fragments in a `<details><summary>` by default in log output
  - [ ] ... or maybe just show fragment IDs but not fragments? Would make `llm logs` output easier to visually scan. Could have a `-e/--expand` option to expand those fragments.
  - [ ] Tempted to add magic that says if the source ended in `.py` or `.js` then it gets wrapped in a syntax highlight block with the correct language tag - bit of a weird feature, maybe have this turned on by `llm logs --syntax` or similar? Might be easier to ignore programming languages but still wrap in triple backticks if it looks like code - if most of the lines are less than 120 chars for example.
- [ ] `llm fragments show hash-or-alias` - to show the full fragment based on the hash from the previous command
- [ ] Comprehensive tests

## Stretch goals

- [ ] Add this to `llm chat` - so you can do `llm chat -f docs`
- [ ] Does this need documentation in the Python docs?
- [ ] Consider if this should be added to templates (maybe not?)
- [ ] Consider how this design might be used to implement prompt caching in Claude

### DIFF
--- a/docs/help.md
+++ b/docs/help.md
@@ -118,7 +118,8 @@ Options:
   --at, --attachment-type <TEXT TEXT>...
                                   Attachment with explicit mimetype
   -o, --option <TEXT TEXT>...     key/value options for the model
-  -f, --fragment TEXT             Fragment to add to prompt
+  -f, --fragment TEXT             Fragment (alias, URL or file path) to add to
+                                  the prompt
   --sf, --system-fragment TEXT    Fragment to add to system prompt
   -t, --template TEXT             Template to use
   -p, --param <TEXT TEXT>...      Parameters for template
--- a/docs/help.md
+++ b/docs/help.md
@@ -118,8 +118,8 @@ Options:
   --at, --attachment-type <TEXT TEXT>...
                                   Attachment with explicit mimetype
   -o, --option <TEXT TEXT>...     key/value options for the model
-  -f, --fragment TEXT             Fragment (alias, URL or file path) to add to
-                                  the prompt
+  -f, --fragment TEXT             Fragment (alias, URL, hash or file path) to
+                                  add to the prompt
   --sf, --system-fragment TEXT    Fragment to add to system prompt
   -t, --template TEXT             Template to use
   -p, --param <TEXT TEXT>...      Parameters for template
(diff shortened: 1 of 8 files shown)


simonw:
Here's a but of a weird problem. We're storing `fragments` in the `logs.db` database - but that means this logic here isn't quite right:

https://github.com/simonw/llm/blob/9b2d4d6256eb82a6a3f67d0e6300f8557778a4c9/llm/cli.py#L476-L482

Here we are only even creating that `logs.db` database at the very end of `prompt()` if the user has logging enabled.

But now we need that database at the start of that CLI command in case we need to resolve any fragment aliases using it.

It's a bit weird that it's called `logs.db` now that it's supporting an additional feature that's not related to if logs are turned on or not. I think I can live with that though.

simonw:
Lots more fragments available now - I added Datasette, `sqlite-utils`, `s3-credentials` and `shot-scraper` docs to https://github.com/simonw/llm-docs

simonw:
Docs: https://github.com/simonw/llm/blob/30331e9a4b24d2f871d71e27df214a4b8233308d/docs/usage.md#fragments

simonw:
Claude idea: since fragments are visible on the `Prompt` class the Anthropic plugin could add a `-o cache 1` option which adds a cache marker directly after the fragments but before the next part of the prompt.

This is the simplest thing that could possibly work, but I think it would be worthwhile.

simonw:
Abandoning this in favor of new PR:
- #859

PR #648, opened 2024-11-23
Update directory.md

vagos:
I had previously added both plugins to `directory.md` (#535), but the commit somehow got lost.

Re-adding them here.


vagos:
@simonw Any idea why the tests are failing?

PR #650, opened 2024-11-26
Add `extra_body` option to SharedOptions for OpenAI

ar-jan:
The OpenAI classes all have an `extra_body` parameter:

> extra_body: Add additional JSON properties to the request

It would be helpful to support this for providers compatible with the OpenAI API who use additional properties.

With this I can use for example `venice_parameters` with [llm-venice](https://github.com/ar-jan/llm-venice):

> llm -m venice/llama-3.1-405b -o extra_body '{"venice_parameters": { "include_venice_system_prompt": true }}' "Repeat the above prompt"

### DIFF
--- a/llm/default_plugins/openai_models.py
+++ b/llm/default_plugins/openai_models.py
@@ -257,6 +257,29 @@ class SharedOptions(llm.Options):
         description="Integer seed to attempt to sample deterministically",
         default=None,
     )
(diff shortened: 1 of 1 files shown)


simonw:
Is this PR complete? I see the bit where the `extra_body` is parsed and validated, but it doesn't seem to be added to the request anywhere?

PR #654, opened 2024-12-02
Register full embedding model names

web-sst:
This update registers the default plugin modules under their full names.
It allows use of the same embedding model names available in ttok.
Aliases for the short names are added for backward compatibility.


simonw:
I think this is an improvement, thanks.

PR #687, opened 2024-12-29
Show the default model when listing all available models.

amjith:
Show the current default model in the list of models in the list printed by `llm models`


amjith:
Closing this in favor of #688. Exact same changes but the PR is made from my `main` branch. Somehow that fixes the failure in the run-cog GH action.

PR #688, opened 2024-12-30
Show the default model when listing all available models.

amjith:
Show the current default model in the list of models in the list printed by `llm models`.


simonw:
Thanks!

PR #689, opened 2024-12-30
Add `--input-style` feature for `prompt`

akx:
This adds `--input-style` for `llm prompt`, which lets the operator choose how input from stdin is passed to the prompt.

For `llm prompt "describe this"` and `input from stdin` as the input, the final prompts would be:

### `--input-style=prepend` (the default, and the current behavior)

`input from stdin describe this`

### `--input-style=append`

`describe this input from stdin`

### `--input-style=fence-prepend`

````
```
input from stdin
```

describe this
````

### `--input-style=fence-append`

````
describe this

```
input from stdin
```
````

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -50,7 +50,8 @@ Will run a prompt of:
 ```
 
 You can change this with the `--input-style` argument; it supports the values `prepend` (the default), `append` (the opposite of `prepend`),
-and `fence-prepend` and `fence-append`, which will fence the standard input content with triple backticks before prepending or appending it.
+`fence-prepend` and `fence-append`, which will fence the standard input content with triple backticks before prepending or appending it, and
+finally `both` and `fence-both`, which could be useful with certain long-context models like GPT-4.1.
 
 For models that support them, {ref}`system prompts <usage-system-prompts>` are a better tool for this kind of prompting.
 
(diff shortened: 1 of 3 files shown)


akx:
Gentle review nudge, @simonw? 😅

Grynn:
Why would you need this? 

I mean you can do:
`(cat somefile; echo "describe this") | llm`

or
`(echo "describe this" ; cat somefile) | llm` 

or
`(echo "describe this\n\n"'```' ; cat summarize.sh ; echo '```') | llm`  if you really want.

akx:
@Grynn Well, all of those are much less convenient to type and iterate upon than this option.

akx:
This just became even more relevant with [this section of the GPT 4.1 prompting guide](https://cookbook.openai.com/examples/gpt4-1_prompting_guide#prompt-organization), emphasis mine:

> Especially in long context usage, placement of instructions and context can impact performance. If you have long context in your prompt, **ideally place your instructions at both the beginning and end of the provided context, as we found this to perform better than only above or below. If you’d prefer to only have your instructions once, then above the provided context works better than below.**


I rebased this and added `both` and `fence-both`.

cc @simonw :)

akx:
rebased.

PR #690, opened 2024-12-30
Add installation docs with uv

ajmarcus:
Add documentation on how to install the llm cli using uv


simonw:
Thanks for this. I'm going to mention it on https://llm.datasette.io/en/stable/setup.html as well and add a section about using `uvx`.

PR #691, opened 2025-01-02
Enable optional experimental TUI via Trogon

sullivancolin:
This PR adds the optional ability use a TUI via the [Trogon project](https://github.com/Textualize/trogon).  It follows the same pattern as [`sqlite_utils`](https://sqlite-utils.datasette.io/en/stable/index.html) by making it an optional dependency.

* `trogon` can be installed either via `llm install trogon` or during the installation of llm itself as in `pipx install 'llm[tui]'`.

* Using `cog` automatically adds the `tui` subcommand to the documentation.  An additional section is added to `usage.md` mimicking the language used in `sqlite_utils`.  There isn't currently any screen shot as part of the documentation like there is with `sqlite_utils`.

* The `cog.yml` and `test.yml` github action workflows now both also install `trogon` as well.

### Sample Screenshot
Screenshot 2025-01-01 at 8 21 13 PM

### DIFF
--- a/.github/workflows/cog.yml
+++ b/.github/workflows/cog.yml
@@ -23,7 +23,7 @@ jobs:
           python-version: '3.11'
 
       - name: Install dependencies
-        run: pip install -e '.[test]'
+        run: pip install -e '.[test,tui]'
 
       - name: Run cog
         run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,6 +25,9 @@ jobs:
       run: |
         pip install -e '.[test]'
         pip install 'pydantic${{ matrix.pydantic }}'
+    - name: Optionally install tui dependencies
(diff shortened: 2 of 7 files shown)


simonw:
I think this would work better as a plugin - so you could run e.g. `llm install llm-tui` and get this functionality that way.

The only reason I didn't do that for `sqlite-utils` is that I didn't have the CLI plugin mechanism setup when I started experimenting with Trogon. I may even remove the feature from `sqlite-utils` and turn it into a plugin now that it's possible to add command plugins there.

simonw:
I made that change to `sqlite-utils` here:

- https://github.com/simonw/sqlite-utils/issues/648

sullivancolin:
Makes sense!  Thanks for the feedback and excellent reference implementation!

Issue #681, opened 2024-12-19
-x/--extract option for returning content of first fenced code block


simonw:
Thought of this after writing up my one-shot prompting tricks here: https://simonwillison.net/2024/Dec/19/one-shot-python-tools/

Idea is to be able to do this:

```bash
llm 'Python CLI tool for finding all files matching an expression, recursively' --extract
```
This would run the provided prompt and return just the content of the first fenced code block (`` ```python `` ...) in the response.

If no fenced code blocks found, returns the whole string.

simonw:
Used `o1-mini` for the initial code: https://chatgpt.com/share/67642b13-72a4-8006-8620-f75e62706554

simonw:
Updated documentation: https://github.com/simonw/llm/blob/67d4a9964501071b3d810a76598ea27cc741ec6f/docs/usage.md#extracting-fenced-code-blocks

simonw:
Demo:
```bash
llm -m gpt-4o-mini 'javascript function to reverse a string' -x
```
Output:
```
function reverseString(str) {
    return str.split('').reverse().join('');
}

// Example usage
const originalString = "Hello, World!";
const reversedString = reverseString(originalString);
console.log(reversedString); // Output: "!dlroW ,olleH"
```
But running `llm logs -c` shows the full prompt and response: https://gist.github.com/simonw/0964cdd9d7e3cb46932d9b017083f25a

simonw:
Templates should support this too.

simonw:
From my blog entry: https://simonwillison.net/2024/Dec/19/one-shot-python-tools/

```bash
llm -s 'You write Python tools as single files. They always start with this comment:

# /// script
# requires-python = ">=3.12"
# ///

These files can include dependencies on libraries such as Click. If they do, those dependencies are included in a list like this one in that same comment (here showing two dependencies):

# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "click",
#     "sqlite-utils",
# ]
# ///' --save python --extract
```

simonw:
I tried this:
```bash
llm -t python 'starlette server providing /?url= API to extract just text using BeautifulSoup'
```
And got back this:
```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "starlette",
#     "httpx",
#     "beautifulsoup4",
# ]
# ///

from starlette.applications import Starlette
from starlette.responses import JSONResponse
from starlette.requests import Request
import httpx
from bs4 import BeautifulSoup

[...]

PR #693, opened 2025-01-02
llm logs -x/--extract option

csabahenk:


simonw:
This is a really good idea - relates to this recently added feature for `llm prompt`:
- #681

PR #695, opened 2025-01-03
Add codespell support (config, workflow to detect/not fix)

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

ATM no typos were found/fixed, and some of the recent commits actually used codespell to find/fix them

- 50520c7c1cb4e3f9353a96cb33cc8b9d18fc0ef8

so I decided it might still be wortwhile proposing just adding config + CI for it.

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,25 @@
+# Codespell configuration is within .codespellrc
+---
+name: Codespell
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+permissions:
+  contents: read
+
+jobs:
+  codespell:
+    name: Check for spelling errors
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Annotate locations with typos
+        uses: codespell-project/codespell-problem-matcher@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/codespell.yml
```diff
@@ -0,0 +1,25 @@
...
+jobs:
+  codespell:
+    name: Check for spelling errors
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Annotate locations with typos
+        uses: codespell-project/codespell-problem-matcher@v1
```
simonw, line 23: I'm a bit nervous about running actions like this from other projects, purely because if that project gets compromised my own CI runs could end up running untrusted code.

My preferred pattern for codespell is to run it like this: https://github.com/simonw/datasette/blob/main/.github/workflows/spellcheck.yml

yarikoptic: I hear you. FWIW

```
permissions:
  contents: read
```

above should at least make it unable to push... as as secrets are also not exposed implicitly, should be safe?? (I am really curious now too ;) )

PR #698, opened 2025-01-07
doc(plugin): adding AmazonBedrock

watany-dev:
By keeping this wonderful plugin in the directory, we can prevent the kind of re-implementation I had to do today.
https://github.com/watany-dev/llm-bedrock-nova


simonw:
Thanks, I really should have done this already!

Issue #717, opened 2025-01-24
`--xl/--extract-last` option


simonw:
Currently the `-x/--extract` option returns the first fenced code block. [Dirk Loss](https://twitter.com/dloss/status/1882845325083496651):

> Why the first code block and not the last? Don't some models self-correct their output sometimes?

simonw:
Will complete this in the PR:
- #718

simonw:
As an experiment I asked o1 to review the diff for me:

```bash
git diff | llm -m o1 'did I miss anything in adding this feature?'
```

# 2025-01-24T17:54:26    conversation: 01jjcq8q9ety0wxa6scdcmgz2a

Model: **o1**

## Prompt:

<details><summary>git diff</summary>

```diff
diff --git a/llm/cli.py b/llm/cli.py
index 1bde74b..606ff3b 100644
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -37,7 +37,7 @@ from .utils import (
     mimetype_from_path,
     mimetype_from_string,
     token_usage_string,
-    extract_first_fenced_code_block,
+    extract_fenced_code_block,
 )
 import base64
 import httpx
@@ -210,6 +210,13 @@ def cli():
 @click.option("async_", "--async", is_flag=True, help="Run prompt asynchronously")
 @click.option("-u", "--usage", is_flag=True, help="Show token usage")
 @click.option("-x", "--extract", is_flag=True, help="Extract first fenced code block")
+@click.option(
+    "extract_last",
+    "--xl",
+    "--extract-last",
+    is_flag=True,
+    help="Extract last fenced code block",
+)
 def prompt(
     prompt,
     system,
@@ -229,6 +236,7 @@ def prompt(
     async_,
     usage,
     extract,
+    extract_last,
 ):
     """
     Execute a prompt
@@ -318,6 +326,8 @@ def prompt(
             to_save["defaults"] = dict(param)
         if extract:
             to_save["extract"] = True
+        if extract_last:
+            to_save["extract_last"] = True
         path.write_text(
             yaml.dump(
                 to_save,
@@ -335,6 +345,7 @@ def prompt(
             raise click.ClickException("Cannot use -t/--template and --system together")
         template_obj = load_template(template)
         extract = template_obj.extract
+        extract_last = template_obj.extract_last
         prompt = read_prompt()
         try:
             prompt, system = template_obj.evaluate(prompt, params)
@@ -343,7 +354,7 @@ def prompt(
[...]

PR #718, opened 2025-01-24
--xl/--extract-last

simonw:
Refs: 
- #717

### DIFF
--- a/docs/help.md
+++ b/docs/help.md
@@ -129,6 +129,7 @@ Options:
   --async                         Run prompt asynchronously
   -u, --usage                     Show token usage
   -x, --extract                   Extract first fenced code block
+  --xl, --extract-last            Extract last fenced code block
   --help                          Show this message and exit.
 ```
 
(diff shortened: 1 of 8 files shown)


simonw:
Found a weird bug with this input:
````
Here are five ways to print a string in Python:

Using the `print()` function is the most straightforward way. Just pass the string as an argument:

```python
print("Hello, World!")
```

You can also use formatted strings (f-strings), which allow you to embed expressions inside string literals:

```python
name = "Alice"
print(f"Hello, {name}!")
```

Another option is using the `format()` method, which inserts variables into a string placeholder:

```python
greeting = "world"
print("Hello, {}!".format(greeting))
```

You can also leverage the old-style string formatting with the `%` operator:

```python
age = 30
print("I am %d years old." % age)
```

Lastly, the `sys` module can be used to print strings to standard output with `sys.stdout.write()`, which does not automatically add a newline:

```python
import sys
sys.stdout.write("Hello, World!\n")
``` 

These methods give you various ways to display strings in Python, depending on your needs.
````
The `last=True` one was not getting that last block.

It turns out there's an invisible single space character after the closing triple backticks there!

PR #751, opened 2025-02-14
Add llm-azure to the plugins directory

kblissett:
[`llm-azure`](https://github.com/fabge/llm-azure/) by fabge is missing from the plugins directory. My understanding is that this plugin is currently the best way to use `llm` with Azure deployed OpenAI endpoints.


rcarmo:
I came here looking for this since I could not fathom why `llm-azure` wasn't listed. Paging @simonw

PR #763, opened 2025-02-21
Add llm-vertex

gmolau:
llm-vertex makes it possible to use Gemini models via GCP.

Related discussion: https://github.com/simonw/llm-gemini/issues/8


gmolau:
@simonw Is this good to merge?

Issue #520, opened 2024-06-24
Fully migrate to Pydantic v2


achille:
After updating to the latest version llm commandline fails
with: ImportError: cannot import name 'field_validator' from 'pydantic

Log: 

% pip install llm
Requirement already satisfied: llm in /opt/homebrew/lib/python3.10/site-packages (0.14)
Requirement already satisfied: click in /opt/homebrew/lib/python3.10/site-packages (from llm) (8.1.3)
Requirement already satisfied: openai>=1.0 in /opt/homebrew/lib/python3.10/site-packages (from llm) (1.30.5)
Requirement already satisfied: click-default-group>=1.2.3 in /opt/homebrew/lib/python3.10/site-packages (from llm) (1.2.4)
Requirement already satisfied: sqlite-utils>=3.35.0 in /opt/homebrew/lib/python3.10/site-packages (from llm) (3.36)
Requirement already satisfied: sqlite-migrate>=0.1a2 in /opt/homebrew/lib/python3.10/site-packages (from llm) (0.1b0)
Requirement already satisfied: pydantic>=1.10.2 in /opt/homebrew/lib/python3.10/site-packages (from llm) (1.10.6)
Requirement already satisfied: PyYAML in /opt/homebrew/lib/python3.10/site-packages (from llm) (6.0)
Requirement already satisfied: pluggy in /opt/homebrew/lib/python3.10/site-packages (from llm) (1.3.0)
Requirement already satisfied: python-ulid in /opt/homebrew/lib/python3.10/site-packages (from llm) (2.2.0)
Requirement already satisfied: setuptools in /opt/homebrew/lib/python3.10/site-packages (from llm) (69.2.0)
Requirement already satisfied: pip in /opt/homebrew/lib/python3.10/site-packages (from llm) (24.0)
Requirement already satisfied: anyio<5,>=3.5.0 in /opt/homebrew/lib/python3.10/site-packages (from openai>=1.0->llm) (3.6.2)
Requirement already satisfied: distro<2,>=1.7.0 in /opt/homebrew/lib/python3.10/site-packages (from openai>=1.0->llm) (1.9.0)
Requirement already satisfied: httpx<1,>=0.23.0 in /opt/homebrew/lib/python3.10/site-packages (from openai>=1.0->llm) (0.24.1)
Requirement already satisfied: sniffio in /opt/homebrew/lib/python3.10/site-packages (from openai>=1.0->llm) (1.3.0)
[...]

PR #775, opened 2025-02-26
Upgrade to Pydantic v2

simonw:
Refs:
- #520

### DIFF
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -1197,9 +1197,9 @@ def models_list(options, async_, query):
             model_with_aliases.model if not async_ else model_with_aliases.async_model
         )
         output = str(model) + extra
-        if options and model.Options.schema()["properties"]:
+        if options and model.Options.model_json_schema()["properties"]:
             output += "\n  Options:"
-            for name, field in model.Options.schema()["properties"].items():
+            for name, field in model.Options.model_json_schema()["properties"].items():
                 any_of = field.get("anyOf")
                 if any_of is None:
                     any_of = [{"type": field.get("type", "str")}]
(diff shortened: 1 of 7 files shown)


llm/cli.py
```diff
@@ -2073,7 +2073,7 @@ def get_history(chat_id):
 def render_errors(errors):
     output = []
     for error in errors:
-        output.append(", ".join(error["loc"]))
```
simonw, line 2076: Was this change necessary?

llm/templates.py
```diff
@@ -13,8 +13,7 @@ class Template(BaseModel):
     extract: Optional[bool] = None
     extract_last: Optional[bool] = None
 
-    class Config:
```
simonw, line 16: This is probably the biggest change, I should review this.

llm/cli.py
simonw: Here's where that's used: https://github.com/simonw/llm/blob/e46cb7e7618d046475c5ddb1586710358f19dbd9/llm/cli.py#L2046-L2051

simonw: It's not neccessary - I tried this:
```pycon
>>> from llm.templates import Template
>>> captured = []
>>> try:
...     Template(invalid=1)
... except Exception as ex:
...     captured.append(ex)
... 
>>> captured[0].errors()
[{'type': 'missing', 'loc': ('name',), 'msg': 'Field required', 'input': {'invalid': 1}, 'url': 'https://errors.pydantic.dev/2.10/v/missing'}, {'type': 'extra_forbidden', 'loc': ('invalid',), 'msg': 'Extra inputs are not permitted', 'input': 1, 'url': 'https://errors.pydantic.dev/2.10/v/extra_forbidden'}]
>>> 
>>> 
>>> def render_errors(errors):
...     output = []
...     for error in errors:
...         output.append(", ".join(error["loc"]))
...         output.append("  " + error["msg"])
...     return "\n".join(output)
... 
>>> 
>>> print(render_errors(captured[0].errors()))
name
  Field required
invalid
  Extra inputs are not permitted
```

PR #819, opened 2025-03-06
Add supports_schema to extra-openai-models.yaml

adaitche:
Recently support for structured output was added. But custom OpenAI-compatible models didn't support the `supports_schema` property in the config file `extra-openai-models.yaml`.

I validated the change by adding the following to my `extra-openai-models.yaml`:
```
- model_id: openai_custom
  model_name: gpt-4o-mini
  api_base: "https://api.openai.com/v1"
  api_key_name: openai
  supports_schema: true
```
and observing the output of
```
uv run llm -m openai_custom --schema 'name,age int,short_bio' 'invent a cool dog'
```

### DIFF
--- a/docs/other-models.md
+++ b/docs/other-models.md
@@ -47,6 +47,8 @@ Add `completion: true` if the model is a completion model that uses a `/completi
 
 If a model does not support streaming, add `can_stream: false` to disable the streaming option.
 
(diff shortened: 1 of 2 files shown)


simonw:
Thanks for this.

PR #836, opened 2025-03-18
-w, --wrap-output: option to wrap long lines of text output

haron:
Screenshot 2025-03-18 at 22 34 20 heic

### DIFF
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -6,6 +6,7 @@ import io
 import json
 import os
 import re
+import textwrap
 from llm import (
     Attachment,
     AsyncConversation,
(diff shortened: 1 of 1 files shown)


simonw:
I think a better way to handle this is with piping to other commands - in particular the `fold -w 80` (TIL about that one) on macOS and Linux (and WSL):

```bash
llm 'lots of otter facts as paragraphs' | fold -w 40
```
A neat thing about that is it works with both streaming and non-streaming responses, and works with `llm logs` as well. Streams are output a line at a time like this:

```
Sure! Here are several paragraphs filled
 with interesting facts about otters:

### Social Structure
Otters are highly social animals, often 
found living in groups known as rafts. A
 typical raft can consist of anywhere fr
om two to twenty otters, though larger g
roups have been observed. Rafts provide 
...
```

Issue #602, opened 2024-11-03
Ability to configure attachment support for models in extra-openai-models.yaml


NightMachinery:
In `~/Library/Application Support/io.datasette.llm/extra-openai-models.yaml`, I have this:
```
- model_id: or:c35s
  model_name: anthropic/claude-3.5-sonnet:beta
  api_base: "https://openrouter.ai/api/v1"
  api_key_name: openrouter
```

When I try to use `-a some_image.png`, I get:
```
Error: This model does not support attachments, but some were provided
```

```
❯ llm --version
llm, version 0.17.1
```

PS: It works with `gemini-1.5-pro-latest` accessed directly from Google. The problem is probably with the OpenRouter indirection.

simonw:
From #593:

> I added custom gpt-4o azure model through `extra-openai-models.yaml` and Api proxy config is
> 
> ```
> - model_id: 4o-azure
>   model_name: azure-gpt4o
>   api_base: "http://10.41.9.156:4000/"
>   api_key_name: azure-gpt4o
> ```
> 
> then command returns
> 
> ```
> "describe this image" \
>       -a IMG_20240908_161441.jpg
> Error: This model does not support attachments, but some were provided
> ```
> 
> Is there option to extra models that indicate model supports attachments or any other function?

PR #843, opened 2025-03-22
Support vision and audio for extra-openai-models.yaml

giuli007:
Should close https://github.com/simonw/llm/issues/602

`extra-openai-models.yaml` can be used to access OpenAI-compatible models, in my case I use it to access models through a proxy. Certain options supported by OpenAI models when used directly are not supported when using this method.

I noticed something similar for schemas has been merged super-recently https://github.com/simonw/llm/pull/819 so I thought to also do the same for other options, namely vision and audio.
This adds support for OpenAI compatible models specified via the `extra-openai-models.yaml` so that models that support it can receive image and audio attachments

---

tested with
```
- model_id: gw-gpt-4o
  model_name: gpt-4o
  api_base: "http://localhost:8888/v1/"
  vision: true
```
and running
`llm -m gw-gpt-4o "describe this image" -a https://static.simonwillison.net/static/2024/pelicans.jpg`

and
```
- model_id: gw-gpt-4o-audio-preview
  model_name: gpt-4o-audio-preview
  api_base: "http://localhost:7777/v1/"
  audio: true
```
and running
`llm -m gw-gpt-4o-audio-preview 'transcript' -a https://static.simonwillison.net/static/2024/video-scraping-pelicans.mp3`

### DIFF
--- a/docs/other-models.md
+++ b/docs/other-models.md
@@ -49,6 +49,10 @@ If a model does not support streaming, add `can_stream: false` to disable the st
 
 If a model supports structured output via JSON schemas, you can add `supports_schema: true` to support this feature.
 
+If a model is a vision model, you can add `vision: true` to support this feature and use image attachments.
+
+If a model is an audio model, you can add `audio: true` to support this feature and use audio attachments.
+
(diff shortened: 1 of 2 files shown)


simonw:
Thanks!

Issue #716, opened 2025-01-24
Design an abstraction for responses that are not just a stream of text


simonw:
LLM currently assumes that all responses from a model come in the form of a stream of text.

This assumption no longer holds!

- Anthropic's new [Citations API](https://www.anthropic.com/news/introducing-citations-api) returns responses that add citation details to some spans of text, [like this](https://gist.github.com/simonw/022d082ccfd636256f72150df344335e).
- DeepSeek Reasoner streams back two types of text - reasoning text and regular text - [as seen here](https://gist.github.com/simonw/a5ca117dd0325c93a5b1f5a18c4a9e34).

And that's just variants of text - multi-modal models need consideration as well. OpenAI have a model that can return snippets of audio already, and models that return images (from OpenAI and Gemini) are coming available very soon too.

simonw:
I had thought that **attachments** would be the way to handle this, but they only work for audio/image outputs - the thing where Claude and DeepSeek can return annotated spans of text feels different.

simonw:
Here's an extract from [that Claude citations example](https://gist.github.com/simonw/022d082ccfd636256f72150df344335e):

```json
{
  "id": "msg_01P3zs4aYz2Baebumm4Fejoi",
  "content": [
    {
      "text": "Based on the document, here are the key trends in AI/LLMs from 2024:\n\n1. Breaking the GPT-4 Barrier:\n",
      "type": "text"
    },
    {
      "citations": [
        {
          "cited_text": "I’m relieved that this has changed completely in the past twelve months. 18 organizations now have models on the Chatbot Arena Leaderboard that rank higher than the original GPT-4 from March 2023 (GPT-4-0314 on the board)—70 models in total.\n\n",
          "document_index": 0,
          "document_title": "My Document",
          "end_char_index": 531,
          "start_char_index": 288,
          "type": "char_location"
        }
      ],
      "text": "The GPT-4 barrier was completely broken, with 18 organizations now having models that rank higher than the original GPT-4 from March 2023, with 70 models in total surpassing it.",
      "type": "text"
    },
    {
      "text": "\n\n2. Increased Context Lengths:\n",
      "type": "text"
    },
    {
      "citations": [
        {
[...]

PR #847, opened 2025-03-23
Annotations

simonw:
- #716

TODO:

- [x] Documentation for `execute()` plugin authors (which is not yet implemented yet)
- [ ] Handle model `.execute()` methods that yield `Chunk` in addition to `str`
- [x] Spin up a dummy model with annotations that I can start writing tests against
- [ ] Implement `Response.chunks()` method [described here](https://github.com/simonw/llm/issues/716#issuecomment-2735156054) for Python API, with docs
- [ ] `llm prompt` should display annotations correctly as they are yielded by the new `.execute()` method
- [ ] Redo the work to add annotation display to `llm logs` - previous prototype [is here](https://github.com/simonw/llm/blob/563a4837f278168dab141d6a8eb3449b6af9b99e/llm/cli.py#L1107-L1360).
- [ ] `llm-openai-plugin` to use annotations for `web_search` tool
- [ ] `llm-anthropic` plugin to use annotations for [Claude citations](https://www.anthropic.com/news/introducing-citations-api)

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -17,12 +17,10 @@ Here's a [YouTube video demo](https://www.youtube.com/watch?v=QUXQNi6jQ30) and [
 Background on this project:
 - [llm, ttok and strip-tags—CLI tools for working with ChatGPT and other LLMs](https://simonwillison.net/2023/May/18/cli-tools-for-llms/)
 - [The LLM CLI tool now supports self-hosted language models via plugins](https://simonwillison.net/2023/Jul/12/llm/)
-- [Accessing Llama 2 from the command-line with the llm-replicate plugin](https://simonwillison.net/2023/Jul/18/accessing-llama-2/)
-- [Run Llama 2 on your own Mac using LLM and Homebrew](https://simonwillison.net/2023/Aug/1/llama-2-mac/)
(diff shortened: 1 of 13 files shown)


simonw:
I hope I got that type stuff right - which cover what happens when you do `for chunk in model.prompt(...)` and the `execute()` method may return `str` or `Chunk` objects. Since `Chunk` has a `__str__()` method I hope things continue to work OK.

simonw:
>  Spin up a dummy model with annotations that I can start writing tests against

Idea: I could have this model class live in the `llm` source code but NOT be registered as a model. Then the tests could register it with a temporary plugin but I could also build an `llm-debug` plugin that, when installed, registers the model for real as well. This would make it easier for me to test things interactively with the `llm` CLI.

simonw:
I have that `llm-debug` plugin now:

```python
import llm
from llm.examples import Markov, AnnotationsModel, AnnotationsModelAsync

@llm.hookimpl
def register_models(register):
    register(Markov())
    register(AnnotationsModel(), AnnotationsModelAsync())
```
Can't package and ship it until I have at least an alpha with `llm/examples.py` in it.

PR #869, opened 2025-03-30
Contribute reference to gtk-llm-chat!

icarito:
I've recently improved it (e.g. added i18n) and set it up to use the python API and load and install as an `llm` plugin. Hope it helps! I have only tested it on Linux.


icarito:
screenshot01
:-)

PR #873, opened 2025-04-01
Should resolve #738, "prompts in python are not saved it the database"

### LINKED ISSUE #738: prompts in python are not saved it the database
Hello!

So I was using llm to make requests using the python api, and it seems that the prompts used in the python api are not saved into the database.
For me it was the expected behavior, because the docs says "llm defaults to logging all prompts and responses to a SQLite database.", and nothing on the python API page nor the Logging to SQLite.
Is this intended and not documented, a bug, or am I missing something?

small reproducible test:
test.py:
```
import llm

model = llm.get_model("gpt-4o-mini")
response = model.prompt("Five surprising names for a pet pelican")
print(response.text())
```

1. `uv init && uv add llm`
2. `uv run llm logs -n 1` -> output your last message
3. `uv run test.py` -> print 5 names
4. `uv run llm logs -n 1` -> should print "Five surprising names for a pet pelican" and the answer, print the same as 2.

Image

Otherwise, thank you very much for your amazing work, it really helps me a lot!

etjones:
adds `log_queries:bool = False` argument to `llm.get_model()`. Default is False, so there's no change to current default behavior, but if `log_queries` is True, all queries will be logged to the sqlite db just as command line queries are.

includes tests. Note that if `log_queries` default were to be changed to True, tests would have to be updated.

All of this code was generated by Claude Sonnet 3.7 in Windsurf, after Windsurf's Google Gemini 2.5 Pro implementation tied itself in knots.

(Note that there are a couple Ruff-generated reorderings of arguments to `Prompt()` calls. Normally, I frown on this kind of messing with somebody else's formatting, but they don't appear to do any harm and do match the argument structure of the Prompt class more closely, so I've left them in. )

### DIFF
--- a/docs/logging.md
+++ b/docs/logging.md
@@ -46,6 +46,16 @@ Number of responses logged:     48
 Database file size:             19.96MB
 ```
 
+(logging-python-api)=
+
+## Logging in Python API
+
+`llm.get_model()` accepts a `log_queries` argument. If `log_queries` is True, 
+all Python API queries for that model will be logged to the database like 
+command-line invocations are. 
+
+Note that the Python API does NOT respect the `llm logs on/off` settings the CLI uses.
+
 (logging-view)=
 
 ## Viewing the logs
--- a/docs/python-api.md
+++ b/docs/python-api.md
@@ -266,6 +266,16 @@ The `response.text()` method described earlier does this for you - it runs throu
 
 If a response has been evaluated, `response.text()` will continue to return the same string.
 
+(python-api-logging-queries)=
+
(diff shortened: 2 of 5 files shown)


jimmy6DOF:
Getting the prompt / response logging is great !!

Just wondering if it will handle conversation , fragments , attachments , and schemas in the logs.db ?

etjones:
As of `llm` v0.12,  it looks like `llm.Response.log_to_db()` accomplishes the goals of this PR-- (but more comprehensively!)

Closing this to reduce the noise

PR #876, opened 2025-04-02
Allow escaping a dollar sign in templates by adding another dollar sign

guspix:
Right now it's impossible to escape a dollar sign on a template, so we can't create a template like the folowing:

```bash
llm 'Does this item cost $10?' --save cost
llm 'What are the contents of my $PATH variable?' --save path
```

This commit allows escaping dollar signs which would allow this:
```bash
llm 'Does this item cost $$10?' --save cost
```

And then we can use the template with:
```bash
llm -t cost "banana"
```

### DIFF
--- a/docs/templates.md
+++ b/docs/templates.md
@@ -32,6 +32,11 @@ Or options:
 ```bash
 llm --system 'Speak in French' -o temperature 1.8 --save wild-french
 ```
+If you want to include a literal `$` sign in your prompt, use `$$` instead:
+```bash
+llm --system 'Estimate the cost in $$ of this: $input' --save estimate
+```
+
 Add `--schema` to bake a {ref}`schema <usage-schemas>` into your template:
 
 ```bash
--- a/llm/templates.py
+++ b/llm/templates.py
@@ -20,7 +20,9 @@ class Template(BaseModel):
     class MissingVariables(Exception):
         pass
 
-    def evaluate(self, input: str, params: Optional[Dict[str, Any]] = None) -> Tuple[Optional[str], Optional[str]]:
+    def evaluate(
+        self, input: str, params: Optional[Dict[str, Any]] = None
+    ) -> Tuple[Optional[str], Optional[str]]:
         params = params or {}
         params["input"] = input
         if self.defaults:
@@ -54,9 +56,15 @@ class Template(BaseModel):
         vars = cls.extract_vars(string_template)
         missing = [p for p in vars if p not in params]
         if missing:
-            raise cls.MissingVariables("Missing variables: {}".format(", ".join(missing)))
+            raise cls.MissingVariables(
+                "Missing variables: {}".format(", ".join(missing))
+            )
         return string_template.substitute(**params)
 
     @staticmethod
     def extract_vars(string_template: string.Template) -> List[str]:
-        return [match.group("named") for match in string_template.pattern.finditer(string_template.template) if match.group("named")]
+        return [
+            match.group("named")
+            for match in string_template.pattern.finditer(string_template.template)
+            if match.group("named")
+        ]
--- a/llm/templates.py
+++ b/llm/templates.py
@@ -20,9 +20,7 @@ class Template(BaseModel):
     class MissingVariables(Exception):
         pass
 
-    def evaluate(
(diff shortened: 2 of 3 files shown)


guspix:
This PR could probably be made much better by documenting this feature, plus providing better error messages when the user passes invalid values like `$10`

simonw:
Got Claude to help me understand the fix. This helped: https://claude.ai/share/ede5044b-f99e-4edc-9270-cc55a2060458

```pycon
>>> import re
>>> import string
>>> 
>>> template = string.Template("This costs $$10, but $item costs $price")
>>> pattern = template.pattern
>>> 
>>> for match in pattern.finditer(template.template):
...     print(f"Match: {match.group(0)}, Named: {match.group('named')}")
... 
Match: $$, Named: None
Match: $item, Named: item
Match: $price, Named: price
```

simonw:
Here's what happens if you don't escape a `$` right now. This works:
```
llm 'Does this item cost $10?' --save cost
```
And can be confirmed with:

    llm templates show cost

```yaml
name: cost
prompt: Does this item cost $10?
```
But if you try to execute it:
```bash
llm -t cost bobcat
```
You get an ugly traceback:
```
Traceback (most recent call last):
  File "/opt/homebrew/Caskroom/miniconda/base/bin/llm", line 8, in <module>
    sys.exit(cli())
  File "/opt/homebrew/Caskroom/miniconda/base/lib/python3.10/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/opt/homebrew/Caskroom/miniconda/base/lib/python3.10/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/opt/homebrew/Caskroom/miniconda/base/lib/python3.10/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/opt/homebrew/Caskroom/miniconda/base/lib/python3.10/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/opt/homebrew/Caskroom/miniconda/base/lib/python3.10/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/opt/homebrew/Caskroom/miniconda/base/lib/python3.10/site-packages/llm/cli.py", line 401, in prompt
    prompt, system = template_obj.evaluate(prompt, params)
  File "/opt/homebrew/Caskroom/miniconda/base/lib/python3.10/site-packages/llm/templates.py", line 37, in evaluate
    prompt = self.interpolate(self.prompt, params)
  File "/opt/homebrew/Caskroom/miniconda/base/lib/python3.10/site-packages/llm/templates.py", line 51, in interpolate
    "Missing variables: {}".format(", ".join(missing))
TypeError: sequence item 0: expected str instance, NoneType found
```

simonw:
The fix works, I agree that adding documentation is important.

guspix:
I should probably have given more info here from my research. The string.Template seems to have four "capturing groups". From the [string Template docs](https://docs.python.org/3/library/string.html#:~:text=escaped%20%E2%80%93%20This%20group,the%20regular%20expression):

> escaped – This group matches the escape sequence, e.g. $$, in the default pattern.
> 
> named – This group matches the unbraced placeholder name; it should not include the delimiter in capturing group.
> 
> braced – This group matches the brace enclosed placeholder name; it should not include either the delimiter or braces in the capturing group.
> 
> invalid – This group matches any other delimiter pattern (usually a single delimiter), and it should appear last in the regular expression.

So in:
```python
         return [ 
              match.group("named") 
              for match in string_template.pattern.finditer(string_template.template) 
             if match.group("named") 
          ]
```
The list comprehension was iterating over all matches of all four types, but the ones that weren't in the "named" group were returning None, and then the code would try to append a string and a None, so it would error out. I think supporting the braced group should probably be simple enough, which would add a bit of flexibility to the templates (think "would three ${product}s fit into a suitcase?", so you can pass "book" and the model would receive the pluralized version). I could look into it if you're interested, though I'm vacationing at the moment so it'd probably be in a bit.

What I think is more important is giving a more descriptive error when there is an invalid expression, like the $10 example, instead of just getting an ugly exception.

Issue #863, opened 2025-03-28
register_fragment_loaders plugin hook (inspired by register_template_loaders)


simonw:
> Similar to:
> - #809 
>
> Then you could do this:
>
> ```bash
> llm -f gh:simonw/llm
> ```
> To pull in a `files-to-prompt` version of the whole repo at once.

_Originally posted by @simonw in https://github.com/simonw/llm/issues/859#issuecomment-2761534692_

simonw:
Should templates that bundle fragments be able to use this `prefix:argument` syntax to tag into fragment loader plugins? Probably?

simonw:
I'm going to build a demo plugin called `llm-fragments-github` which lets you do this:
```bash
llm -f github:simonw/llm 'tell me about this repo'
```
It does a fresh clone of that repo and returns ALL text files as fragments like this (similar but not exactly the same as `files-to-prompt`):
```xml
<document source="README.md">
contents goes here
</document>
```

simonw:
Got Claude to write that plugin: https://claude.ai/share/57c00417-1488-4e68-825b-5c4fe6ab2f8f

simonw:
Whoa it worked!
```bash
llm -f github:simonw/files-to-prompt 'explain this code'
```
Output started:

> The content you've provided appears to be the complete code base for a software project named `files-to-prompt`, as well as some associated documentation, testing scripts, configuration files, and CI/CD workflows.
> 
> ### Overview of the `files-to-prompt` Project
> 
> The `files-to-prompt` project is a command-line tool designed to concatenate multiple files or the contents of directories into a single prompt, which is particularly useful when interacting with large language models (LLMs). Here’s a summary of the key elements:
> 
> [...]

It failed at the end of the response:

```
  File "/Users/simon/.local/share/virtualenvs/llm-p4p8CDpq/lib/python3.10/site-packages/sqlite_utils/db.py", line 533, in execute
    return self.conn.execute(sql, parameters)
sqlite3.IntegrityError: UNIQUE constraint failed: prompt_fragments.response_id, prompt_fragments.fragment_id
```

simonw:
That's this code here https://github.com/simonw/llm/blob/9b78e39d45006d0c7808f8492ec3946f49ef6bd4/llm/models.py#L427-L446

simonw:
Got a debugger running like this:
```bash
LLM_RAISE_ERRORS=1 python -i -m llm -f github:simonw/files-to-prompt 'explain this code'
```
```
[...]

PR #886, opened 2025-04-06
register_fragment_loaders() hook

simonw:
- #863

TODO:

- [x] Plugin hook documentation
- [x] Make `llm.utils.FragmentString` available as `llm.FragmentString`
- [x] Upgrade hook to be able to return a list of `FragmentString` too, document that
- [x] Implement the code that calls the hook
- [x] Write tests that exercise the plugin hook
- [x] Reconsider `FragmentString` - maybe rename to `llm.Fragment` and stop subclassing `str`

Decided to leave out of scope for the moment:

- Consider the Python API for calling fragments provided by a plugin
- Ensure this works with YAML templates too, and `llm --save`

### DIFF
--- a/docs/plugins/plugin-hooks.md
+++ b/docs/plugins/plugin-hooks.md
@@ -115,7 +115,9 @@ Plugins can register new fragment loaders using the `register_template_loaders` 
 
 The `prefix` specifies the loader. The `argument` will be passed to that registered callback..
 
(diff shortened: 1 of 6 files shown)


simonw:
I think `llm logs -f foo -f bar` should be an AND operation, then if you pass `-f dir:llm` (using a fragment plugin that returns multiple fragments) it can search for all prompts that used all of them.

Issue #887, opened 2025-04-06
Mypy is missing stubs/marker


johncalls:
I'm using `llm` as a library. Running mypy on my project results in the following error:

> error: Skipping analyzing "llm": module is installed, but missing library stubs or py.typed marker  [import-untyped]

Adding a file `py.typed` _locally_ to `llm` resolves the error for me. I'd suggest adding the marker (or stubs) to the library.

The corresponding PEP can be found here: https://peps.python.org/pep-0561/.

simonw:
Ran a few rounds with Claude to check the consequences of doing this: https://claude.ai/share/ac09a168-828e-4465-a6f6-6c83511a38af - looks to me like having `py.typed` in the `llm/` folder is already protected by my `mypy` job, which will fail if any of the public methods in that package are not correctly typed: https://github.com/simonw/llm/blob/0cc26b3d4f9efea3946371ca1329f0f5dde1e935/.github/workflows/test.yml#L39-L42

simonw:
I'm doing this first:
- #907

simonw:
I'm trying to figure out if it's worth adding this to `pyproject.toml`:

```toml
[tool.mypy]
disallow_untyped_defs = true
disallow_incomplete_defs = true
```
Is that the default already? I want to be absolutely sure I don't accidentally ship a release that breaks `mypy` for other people who are depending on the new `py.typed` file.

simonw:
OK, this is in a branch now. I'll open a PR and test it there.

simonw:
The branch is now available to install from:

    https://codeload.github.com/simonw/llm/zip/refs/heads/py.typed

simonw:
I think the way to test this is with a new temporary Python project that depends on LLM and uses `mypy`.

simonw:
```bash
cookiecutter gh:simonw/click-app
```
Then:
```
  [1/6] app_name (): cowsay-via-llm
  [2/6] description (): cowsay with an LLM
  [3/6] hyphenated (cowsay-via-llm): 
  [4/6] underscored (cowsay_via_llm): 
  [5/6] github_username (): simonw
  [6/6] author_name (): Simon Willison
```
```bash
cd cowsay-via-llm
uv sync --extra
```
I added `llm` as a dependency to `pyproject.toml` (the current shipped version, not the branch).

This is a _really silly_ demo:

```python
import click
import llm

SYSTEM = """
Render this as ASCII art like cowsay does.
Return plain text no markdown or fenced code block
[...]

PR #909, opened 2025-04-10
py.typed marker file

simonw:
Refs:
- #887


simonw:
Manually tested extensively here and it works as it should:
- https://github.com/simonw/llm/issues/887#issuecomment-2795461824

Issue #772, opened 2025-02-25
Connection error on fresh install when behind proxy - certificate issue


simonfisher-ss:
When trying any model I get:
Error: Connection error.

I'm behind a zscaler proxy/firewall, and there is a HTTPS certificate issue. 

The uv package manager handles this situation well, by [allowing use of the system's native certificate store](https://docs.astral.sh/uv/configuration/authentication/#http-authentication) through the --native-tls command.

Could a similar function be added to llm? 

Thanks

gachikuku:
same issue herw

wallrj-cyberark:
I find that setting the `SSL_CERT_FILE` environment variable is sufficient to get `llm` to use my corporate trust store.
 * https://www.python-httpx.org/environment_variables/


On Fedora, I add the corporate root CA as follows (these are from memory and may be inaccurate):

```
trust anchor --store ~/my-corporate-root.pem
update-ca-trust
```

Which adds the corporate root to `/etc/ssl/cert.pem`.

Then I have to install llm using Python 3.12, for the reasons given here:
 * https://github.com/ansible/ansible-dev-environment/issues/343
 * https://docs.python.org/3/library/ssl.html#ssl.SSLContext.verify_flags

```
$ uv tool install llm -p 3.12
Resolved 32 packages in 726ms
Prepared 1 package in 210ms
Installed 32 packages in 336ms
 + annotated-types==0.7.0
 + anyio==4.11.0
 + certifi==2025.8.3
 + click==8.3.0
 + click-default-group==1.2.4
 + condense-json==0.1.3
 + distro==1.9.0
 + h11==0.16.0
 + httpcore==1.0.9
 + httpx==0.28.1
 + idna==3.10
 + jiter==0.11.0
 + llm==0.27.1
 + openai==1.109.1
 + pip==25.2
 + pluggy==1.6.0
 + puremagic==1.30
 + pydantic==2.11.9
 + pydantic-core==2.33.2
 + python-dateutil==2.9.0.post0
 + python-ulid==3.1.0
 + pyyaml==6.0.3
 + setuptools==80.9.0
 + six==1.17.0
 + sniffio==1.3.1
 + sqlite-fts4==1.0.3
 + sqlite-migrate==0.1b0
 + sqlite-utils==3.38
 + tabulate==0.9.0
 + tqdm==4.67.1
 + typing-extensions==4.15.0
 + typing-inspection==0.4.1
Installed 1 executable: llm
```

[...]

PR #913, opened 2025-04-11
Fix SSL Certificate Validation Behind Corporate Proxies

prateek:
This PR adds support for configuring SSL certificate handling when using LLM behind corporate proxies like Zscaler.

## Problem

As described in issue #772, users behind corporate proxies or firewalls that perform SSL inspection (like Zscaler) encounter connection errors because the HTTPS certificate validation fails.

Unlike tools like `uv` (which has a `--native-tls` option), LLM didn't have a way to configure certificate handling to work in these environments.

## Solution

This PR adds environment variables to configure SSL certificate handling:

```bash
# Use the system's native certificate store (similar to uv's --native-tls)
export LLM_SSL_CONFIG=native_tls
```

```bash
# Or specify a custom CA bundle
export LLM_CA_BUNDLE=/path/to/cert.pem
```

The configuration options include:

- `LLM_SSL_CONFIG=native_tls`: Use the system's native certificate store
- `LLM_SSL_CONFIG=no_verify`: Disable certificate verification (not recommended for production)
- `LLM_CA_BUNDLE=/path/to/cert.pem`: Use a custom CA bundle file

## Implementation Details

- Added a helper function `_configure_ssl_client` that reads environment variables for SSL configuration
- Added validation for SSL configuration values with helpful warning messages
- Added certificate file existence checking to prevent silent failures
- Integrated the helper function with the `get_client` method
- Added comprehensive tests for all scenarios and configurations
- Added detailed documentation in the docs

### DIFF
--- a/llm/default_plugins/openai_models.py
+++ b/llm/default_plugins/openai_models.py
@@ -12,6 +12,7 @@ from enum import Enum
 import httpx
 import openai
 import os
+import warnings
 
 from pydantic import field_validator, Field
 
@@ -535,8 +536,16 @@ class _Shared:
             kwargs["api_key"] = "DUMMY_KEY"
         if self.headers:
             kwargs["default_headers"] = self.headers
+
+        # Configure SSL certificate handling from environment variables
+        ssl_client = _configure_ssl_client(self.model_id)
+
+        if ssl_client:
+            kwargs["http_client"] = ssl_client
+
         if os.environ.get("LLM_OPENAI_SHOW_RESPONSES"):
-            kwargs["http_client"] = logging_client()
+            if "http_client" not in kwargs:
+                kwargs["http_client"] = logging_client()
         if async_:
             return openai.AsyncOpenAI(**kwargs)
         else:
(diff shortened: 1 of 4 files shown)


llm/default_plugins/openai_models.py
```diff
@@ -798,3 +807,54 @@ def redact_data(input_dict):
         for item in input_dict:
             redact_data(item)
     return input_dict
+
+
+def _configure_ssl_client(model_id):
```
akx, line 812: This function doesn't seem to be using the `model_id` argument for anything...

llm/default_plugins/openai_models.py
```diff
@@ -535,8 +536,16 @@ def get_client(self, key, *, async_=False):
...
+        # Configure SSL certificate handling from environment variables
+        ssl_client = _configure_ssl_client(self.model_id)
+
+        if ssl_client:
+            kwargs["http_client"] = ssl_client
+
         if os.environ.get("LLM_OPENAI_SHOW_RESPONSES"):
-            kwargs["http_client"] = logging_client()
+            if "http_client" not in kwargs:
+                kwargs["http_client"] = logging_client()
```
akx, line 548: Doesn't this mean that `LLM_OPENAI_SHOW_RESPONSES` won't work if a custom SSL configuration is in use?

llm/default_plugins/openai_models.py
```diff
@@ -798,3 +807,54 @@ def redact_data(input_dict):
...
+    if ssl_config and ssl_config not in valid_ssl_configs:
+        warnings.warn(
+            f"Invalid ssl_config value: {ssl_config}. Valid values are: {', '.join(valid_ssl_configs)}"
+        )
+        return None
+
+    try:
+        if ssl_config == "native_tls":
+            # Use the system's native certificate store
+            return DefaultHttpxClient(transport=httpx.HTTPTransport(verify=True))
```
bissonex, line 842: I think this will not work. Have a look at https://github.com/encode/httpx/discussions/2490

simonfisher-ss:
Is there an update with this? It doesn't seem clear if the original PR works to enable native TLS...

bissonex:
> Is there an update with this? It doesn't seem clear if the original PR works to enable native TLS...

This draft is working for me: https://github.com/simonw/llm/pull/979 with the only caveat that it needs Python 3.10

prateek:
Sorry, I got caught up in something and didn't circle back. My DAYJOB finally got back to me about ZScaler - they're updating the Root CA in the next ~2w so I won't need this in a few days. 

Going to close it, lets use @bissonex's version if still relevant.

PR #927, opened 2025-04-16
Mention llm-fragments-doc plugin

mbode:
I was really enjoying the new fragments functionality and thought it would be useful being able to work directly with files like pdf, docx, xlsx etc. I built a small prototype using [docling](https://github.com/docling-project/docling) and wanted to politely ask if you'd like to have it mentioned here. Of course, the functionality is rather basic and could be easily integrated into llm itself, but the dependency is somewhat non-light-weight. I'm also open to any suggestions for the plugin.

In any case, thank you very much for all your work!


simonw:
This plugin is great! Would you be OK renaming the fragment to `docling:` just to avoid any confusion between `doc:` and the `docs:` one from https://github.com/simonw/llm-docs?

PR #929, opened 2025-04-17
Add llm-fragments-pypi to plugin directory

samueldg:
As the name suggests, this plugin gets PyPI packages as fragments.
Namely the project metadata and description.

See README for more usage examples:
https://github.com/samueldg/llm-fragments-pypi?tab=readme-ov-file#usage


simonw:
This is a really neat plugin, thanks!

Issue #356, opened 2023-11-26
minor quirks in chat w/ new readline functionality


roblg:
Saw the `readline` addition this week and wanted to check it out. Can confirm that things like arrow keys, alt+delete (on macOS) now work (yay!). But there are some weird interactions between (I think) `readline` and `click` prompts.

Quirks:

* When using 'alt+delete' to delete the last word on a line, the space character of the prompt is also deleted.
  <img width="451" alt="image" src="https://github.com/simonw/llm/assets/777138/da27daac-8df1-47a4-8d6c-74211f9e1846">

* Accessing history with `Ctrl+r` is allowed and functional, but breaking out w/ `Ctrl+c` causes `click` prompt to abort. Exiting with `Esc` deletes the line including the `click` prompt.
  <img width="461" alt="image" src="https://github.com/simonw/llm/assets/777138/8d4d8c00-f28b-42c1-8924-5a1ec4b569c8">

I almost didn't file these because they're so, so nitty, but the `readline` change is fresh so I thought it would be better to do it now than later. 😃 

Also, thanks for the tool and the extensive docs and blogs -- it's been really helpful, and a lot of fun. 🙇‍♂️

yorickvP:
Should be fixed by https://github.com/pallets/click/pull/2969

PR #940, opened 2025-04-20
Use click.visible_prompt_func directly instead of prompt() to fix edi…

### LINKED ISSUE #582: Problem with overwriting in "llm chat" line wrapping
In a "llm chat" prompt (single-line input), whenever I reach the last column of the line, the cursor overwrites a single character and starts confusing where the line begins.

In this example, I show the problem by slowly typing "test" at the end of the line and then typing C-u to go back to the beginning of the line. The "e" in test at the end of the line is gone and then the space between the ">" prompt and where I am typing is gone.

llm

yorickvP:
click.prompt() was manually printing the prompt characters, confusing readline. This was noticable when using ctrl+backspace or when wrapping a line.

Fixes:
- #582 

Partially:
- #356

### DIFF
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -909,7 +909,11 @@ def chat(
     accumulated = []
     end_token = "!end"
     while True:
-        prompt = click.prompt("", prompt_suffix="> " if not in_multi else "")
+        prompt = ""
+        while prompt == "" or prompt.isspace():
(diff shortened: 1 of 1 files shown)


yorickvP:
PR'd upstream instead. https://github.com/pallets/click/pull/2969

PR #954, opened 2025-04-25
Add PDF fragment loader plugin to directory

agustif:
Hi! After making this feature for my other arxiv plugin i thought it could be useful for generic PDF's too! so i made

 **[llm-plugin-pdf](https://github.com/agustif/llm-plugin-pdf)** provides a `-f pdf:` loader that can load local or remote PDF files as fragments.

A little wrapper around pyMuPDF that will try to parse a PDF text and images into markdown to provide a PDF's files contents as a fragment

this should use way less tokens than feeding a full PDF to a model directly, most papers are actually built from source so they have great support that doesn't rely on clunky OCR (i explored using grobid for other uses, but requires a server which made it a nono for this, pyMuPDF worked well on my tests and i was able to also parse the pdf.images into base64 encoded data so it's all passed as fragments to the model, not only text)


simonw:
This plugin can now be upgraded to pass those images as attachments, not as base64 encoded strings:
- #972

PR #955, opened 2025-04-25
Adds arxiv plugin to the plugins directory in docs

agustif:
I made https://github.com/agustif/llm-arxiv to make it easy to load arxiv papers as fragments,it takes either a full url or just arXiv id like `-f arxiv:2504.17544v1`


agustif:
fixed conflicts would love to get this merged @simonw

PR #957, opened 2025-04-25
Update doc for extra-openai-models.yaml, include mandatory model_name

rahimnathwani:
The documentation doesn't work as-is: if you create a file without model_name, then no calls to llm will work.


simonw:
Thanks for catching this.

PR #965, opened 2025-04-29
Add gpt-4.1 model family to default models

abizer:
hopefully this was done correctly

```bash
.venv > llm -m gpt-4.1-mini "hello world"
Hello! How can I assist you today?
```

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,5 +1,9 @@
 # Changelog
 
+(v0_25_1)=
+## 0.25.1 (2025-04-29)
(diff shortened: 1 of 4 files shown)


simonw:
I was going to leave this to the new [llm-openai](https://github.com/simonw/llm-openai-plugin) plugin but actually I think adding it to core as well makes more sense, at least until I entirely remove OpenAI models from core in favor of that plugin.

simonw:
That test failure is a frustrating intermittent GitHub error, nothing to do with this change.

PR #969, opened 2025-05-02
Add '!edit' command to modify prompts in chat sessions using default editor

Hopiu:
## Description

This pull request introduces a new feature to the chat command in the CLI: the ability to edit the current prompt using the user's default editor by typing `!edit` at the prompt. When a user enters !edit, the CLI opens the default editor (as configured by the `$EDITOR` environment variable) for prompt editing. If the editor is closed without saving, the user is notified and prompted again. If the prompt is saved and not empty, it is used as the next message in the chat session.

## Motivation
This feature improves the chat experience by allowing users to compose or revise prompts in their preferred text editor, which is especially useful for longer or more complex inputs.


simonw:
This is a _really_ smart feature, thanks! I'm going to test it manually myself as well because writing automated tests for this is pretty tricky, and Click already has extensive tests of its own.

simonw:
... I tested it, it works great!

docs/changelog.md
```diff
@@ -419,6 +419,7 @@ This starts a chat session:
 ```
 Type 'exit' or 'quit' to exit
+Type '!edit' to open your default editor and modify the prompt.
```
simonw, line 422: I'm going to drop this from the changelog as I'm OK with that showing how the feature looked back in version 0.10.

simonw:
That cog test failure is a frustrating intermittent GitHub error, nothing to do with this change.

PR #970, opened 2025-05-03
Honor default model options when using aliases

### LINKED ISSUE #968: Models accessed via an alias do not load their default options
```bash
llm models options 
```
```
mlx-community/Qwen3-8B-4bit:
  unlimited: 1
  max_tokens: 32768
```
But when I run a prompt and then `llm logs -c --json` I get this:
```json
    "options_json": {},
```
That could be a bug, but I think it means we are not correctly recording any default options in the SQLite logs.

david-jarman:
When getting the default options for a model, the resolved model_id was not being used if the user invoked the cli using an alias, like `llm -m <alias> "..."`.

Fixes #968

### DIFF
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -667,7 +667,7 @@ def prompt(
             raise click.ClickException(render_errors(ex.errors()))
 
     # Add on any default model options
-    default_options = get_model_options(model_id)
+    default_options = get_model_options(model.model_id)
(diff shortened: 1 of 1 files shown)


simonw:
Thanks for this! I fixed it in this commit, should have looked at the PRs first: 16770611ca280536062b64849c41cfc673cf8fca

Issue #972, opened 2025-05-04
Allow fragment loader plugins to return a mixture of fragments and attachments


simonw:
It would be useful if a `-f prefix:value` plugin could return attachments in addition to fragments of text.

Example of a plugin that could use this is `llm-arxiv` which could then return a mixture of text content and attachments for the images from the paper:

- https://github.com/agustif/llm-arxiv/issues/2

simonw:
System fragments should not allow this because it doesn't make sense to attempt to put an attachment in a system prompt.

simonw:
I guess I'd change the signature of the fragment loader plugin mechanism to allow it to return a list of combined Fragment and Attachments.

https://github.com/simonw/llm/blob/07ccfbcee5479f7a4b2f35935b817aebed9502a4/llm/__init__.py#L123-L127

simonw:
Before I commit to it, is this definitely a good feature idea?

Some things this could do:

- `-f arxv:paper-ref` returning both text and images from a paper
- `-f pdf:file.pdf` could be a plugin that extracts text and images from a PDF, for models that cannot handle PDF data yet
- `-f dalle:prompt-goes-here` could call an image generator and feed in the image. Not sure why you would do this but could be fun I guess?
- `-f pdf-split:file.pdf` could split a PDF file into one image per page. Might be better as a template loader plugin though?

That's really the crux of this: should this not be part of fragments at all since template loaders can do it already?

simonw:
The difference between template loaders and fragment loaders is clearly pretty confusing right now.

simonw:
The big difference between templates and fragments is that you can provide more than one fragment. That's actually a pretty compelling argument for `-f` to be allowed to return images, especially as context lengths keep on getting longer.

`-f video-frames:video.mp4?fps=2` could be a plugin that uses `ffmpeg` or similar to turn a video into images (2 per second).

simonw:
I built a protoype and got o4-mini to write me the first version of that plugin, which works _extremely_ well:
```python
import os
import tempfile
import subprocess
import glob
from urllib.parse import urlparse, parse_qs
import llm


@llm.hookimpl
def register_fragment_loaders(register):
    register("video-frames", video_frames_loader)


def video_frames_loader(argument: str):
    """
    Fragment loader "video-frames:<path>?fps=N&timestamps=1"
    - extracts frames at `fps` per second (default 1)
    - if `timestamps=1`, overlays "filename hh:mm:ss" at bottom-right
    """
[...]

PR #974, opened 2025-05-04
Fragment plugins can now optionally return attachments

simonw:
- #972

Still needs:

- [x] `mypy` fixes
- [x] Tests
- [x] Documentation
- [x] Example plugin
  - [x] Prototype in https://github.com/simonw/llm/issues/972#issuecomment-2849412037
  - [ ] Needs a release

### DIFF
--- a/docs/fragments.md
+++ b/docs/fragments.md
@@ -143,7 +143,9 @@ This plugin turns a single call to `-f github:simonw/s3-credentials` into multip
 
 Running `llm logs -c` will show that this prompt incorporated 26 fragments, one for each file.
 
-Running `llm logs -c --usage --expand` includes token usage information and turns each fragment ID into a full copy of that file. [Here's the output of that command](https://gist.github.com/simonw/c9bbbc5f6560b01f4b7882ac0194fb25).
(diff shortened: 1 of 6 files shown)


simonw:
```
  mypy
llm/cli.py:142: error: Argument 1 to "extend" of "list" has incompatible type "list[Fragment | Attachment]"; expected "Iterable[Fragment]"  [arg-type]
llm/cli.py:159: error: Incompatible return value type (got "list[Fragment]", expected "list[Fragment | Attachment]")  [return-value]
llm/cli.py:159: note: "List" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
llm/cli.py:159: note: Consider using "Sequence" instead, which is covariant
llm/cli.py:159: note: Perhaps you need a type annotation for "resolved"? Suggestion: "list[Fragment | Attachment]"
Found 2 errors in 1 file (checked 16 source files)
error: Recipe `lint` failed on line 22 with exit code 1
```

simonw:
[TIL (via o4-mini)](https://gist.github.com/simonw/fac1e5a0c3d0dbc01558e8f990824934):

> This is just a classic case of mypy assuming that
> 
> ```python
> resolved = []
> ```
> 
> is a `List[Fragment]` (since you first append `Fragment`-only) and then choking when you do
> 
> ```python
> resolved.extend(result)  # result: List[Fragment|Attachment]
> ```
> 
> To fix it, give `resolved` an explicit annotation so that mypy knows it really holds both `Fragment` and `Attachment`: [...]
> 
> ```python
> resolved: List[Union[Fragment, Attachment]]
> ```

PR #985, opened 2025-05-10
Load model options in chat command

indirect:
This change ensures the `chat` command will load a model's default options. Having default options work for `prompt` but get ignored by `chat` is a very surprising situation, so this is my attempt to resolve that issue.


simonw:
This is a good catch - thanks for updating the tests too.

PR #1050, opened 2025-05-19
Allow prompting without quotes

singiamtel:
Before

```bash
$ llm hi hows it going
Usage: llm prompt [OPTIONS] [PROMPT]
Try 'llm prompt --help' for help.

Error: Got unexpected extra arguments (hows it going)
```

After

```bash
$ uv run llm hi hows it going
Hello! I'm here and ready to help. How can I assist you today?
```

It doesn't break stdin input, if that's a concern

```bash
$ echo "hello hows it going" | uv run llm
Hello! I'm doing well, thank you. How about you? How can I assist you today?
```

### DIFF
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -302,7 +302,7 @@ def cli():
 
 
 @cli.command(name="prompt")
-@click.argument("prompt", required=False)
+@click.argument("prompt", required=False, nargs=-1)
 @click.option("-s", "--system", help="System prompt to use")
 @click.option("model_id", "-m", "--model", help="Model to use", envvar="LLM_MODEL")
 @click.option(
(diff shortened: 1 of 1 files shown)


singiamtel:
For the red tests #1051. No idea what might be going on with the `run-cog` one though

simonw:
I've thought about this before, but there's one nasty usability issue that it introduces.

Right now if you typo one of the many LLM sub-commands you usually (but not always) get an error:

```bash
llm embed-multip --files docs/ '*.md'
```
Output:
```
Usage: llm prompt [OPTIONS] [PROMPT]
Try 'llm prompt --help' for help.

Error: No such option: --files
```
Mis-spelling `embed-multi` caused it to run `llm prompt` but the extra `--files` option was invalid, causing the error.

If `llm prompt` takes multiple arguments like this ALL typos will end up executing that command, no matter what options you sent them.

As such, I'm going to stick to requiring quotes around the prompt.

luciantin:
@singiamtel You can pass the args to a sh file and wrap the args in there, for ex. :

llmc.sh

```
llm --some_extra_flags "$*"
```

Put it in your path and then run it as :
```
llmc Hello There
```

PR #1051, opened 2025-05-19
Fix flaky test

singiamtel:
Seems like a regression in https://github.com/simonw/llm/commit/f879b816e4d5ad03faeaedf5f2a1e11dfe1c31e0


simonw:
Thanks, looks like I fixed that one already here though: https://github.com/simonw/llm/commit/4be2bbdc0efdcd0e5fe2c7fabb4844bb61ec0a10

PR #1068, opened 2025-05-23
Support supports_tools parameters in openai compatible models

mahesh-hegde:
Adds support for `supports_tools` parameter in OpenAI compatible models.


simonw:
Good call, thanks.

PR #1077, opened 2025-05-24
Provide type hint for response_json

ar-jan:
Satisfies the type checker, I got:

> Cannot assign to attribute "response_json" for class "Response"
> Expression of type "dict[str, Any]" cannot be assigned to attribute "response_json" of class "Response"
> "dict[str, Any]" is not assignable to "None"


simonw:
Thanks, looks reasonable to me.

Issue #1089, opened 2025-05-26
Toolbox instantiation should be logged to DB when logging toolbox methods


simonw:
Split from:
- #1086
- #1059

Also refs:
 - #1080

simonw:
Let's use https://github.com/simonw/llm-tools-datasette as an example. It currently has one argument:

```bash
llm --tool 'Datasette("https://datasette.io/content")' "Show tables" --td
```
I'm contemplating adding a second optional one for authentication (a key to lookup in `llm.get_key()):

- https://github.com/simonw/llm-tools-datasette/issues/2

If someone runs the above `llm --tool` prompt what would I need to store in the DB and where would I put it?

If the tool class is instantiated but no methods on it are called should we still record that we instantiated it? The logs need to know "the model had the option to use these tools and chose not to". I think the configuration of the class IS relevant to that, since maybe the model decided not to use tools based on how they had been configured (but can it see that?)

simonw:
Part of the challenge here is where this happens.

CLI users have their toolboxes instantiated for them here: https://github.com/simonw/llm/blob/e23e13e6c7d4e865f0e31927166930051a969e22/llm/cli.py#L807

Or here in `llm chat`:

https://github.com/simonw/llm/blob/e23e13e6c7d4e865f0e31927166930051a969e22/llm/cli.py#L1091

But Python API users instantiate tools directly themselves:

```python
import llm
from llm_tools_datasette import Datasette

model = llm.get_model("gpt-4.1-mini")

result = model.chain(
    "Show most interesting tables",
    tools=[Datasette("https://datasette.io/content")],
).text()
```
How do we record the "this class was instantiated at this moment and with these arguments" information such that, later on, this code in `Response.log_to_db()` can see that information and put it somewhere useful?

https://github.com/simonw/llm/blob/e23e13e6c7d4e865f0e31927166930051a969e22/llm/models.py#L815-L845

simonw:
I'm inclined to have the `llm.Toolbox` subclass do some magic here, somehow recording its moment of instantiation such that later on code with access to `self.prompt.tools` can see what it recorded.

simonw:
The main impact this has is that `llm logs -T Datasette` might not work. A hack would be to teach it to look for `Datasette_*` in the `name` column of the `tools` table, which would actually work because the naming convention for tool classes is *not* to include underscores in them.

simonw:
o4-mini's reply made me think that maybe I could have this table:

```sql
[...]

PR #1098, opened 2025-05-27
Log tool_instances to database

simonw:
Refs:
- #1089

TODO:
- [x] Tests for toolboxes in sync mode
- [x] Tests for async toolbox methods in sync mode (`async def list_files()`)
- [x] Tests for toolboxes in async mode
- [x] Tests for async toolbox methods in async mode

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,6 +186,7 @@ See also [the llm tag](https://simonwillison.net/tags/llm/) on my blog.
   * [Trying out tools](https://llm.datasette.io/en/stable/tools.html#trying-out-tools)
   * [LLM’s implementation of tools](https://llm.datasette.io/en/stable/tools.html#llm-s-implementation-of-tools)
   * [Tips for implementing tools](https://llm.datasette.io/en/stable/tools.html#tips-for-implementing-tools)
+  * [Warning: tools are at risk from malicious attacks](https://llm.datasette.io/en/stable/tools.html#warning-tools-are-at-risk-from-malicious-attacks)
 * [Schemas](https://llm.datasette.io/en/stable/schemas.html)
   * [Schemas tutorial](https://llm.datasette.io/en/stable/schemas.html#schemas-tutorial)
     * [Getting started with dogs](https://llm.datasette.io/en/stable/schemas.html#getting-started-with-dogs)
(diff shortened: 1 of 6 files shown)


simonw:
```bash
llm -f tests/test_plugins.py -f llm/models.py -m o3 -s \
  'The memory_set and memory_get test results look wrong, {"args": [], "kwargs": {}} should not happen, and {"name": "Memory_get", "output": "two", "instance": None} has instance None when it should be populated'
```
https://gist.github.com/simonw/2709b0b9b960fd13edad70b830f3509f - looks like it figured it out for me.

Got Claude to write me some tests to help exercise that: https://claude.ai/share/6153b936-4c50-4af4-a37b-72339bc01717

```python
def test_toolbox_config_capture():
    """Test that Toolbox captures __init__ parameters in _config"""
    
    # Single positional arg
    class Tool1(Toolbox):
        def __init__(self, value):
            pass
    
    assert Tool1(42)._config == {"value": 42}
    
    # Multiple positional args
    class Tool2(Toolbox):
        def __init__(self, a, b, c):
            pass
    
    assert Tool2(1, 2, 3)._config == {"a": 1, "b": 2, "c": 3}
    
    # Keyword args with defaults
    class Tool3(Toolbox):
        def __init__(self, name="default", count=10):
            pass
    
    assert Tool3()._config == {"name": "default", "count": 10}
    assert Tool3(name="custom", count=20)._config == {"name": "custom", "count": 20}
    
    # Mixed args
    class Tool4(Toolbox):
        def __init__(self, required, optional="default"):
            pass
    
    assert Tool4("hello")._config == {"required": "hello", "optional": "default"}
    assert Tool4("world", optional="custom")._config == {"required": "world", "optional": "custom"}
    
    # Var args excluded
    class Tool5(Toolbox):
        def __init__(self, regular, *args, **kwargs):
            pass
    
    assert Tool5("test", 1, 2, extra="value")._config == {"regular": "test"}
    
    # No init
    class Tool6(Toolbox):
        pass
    
    assert Tool6()._config == {}
```
They fail against the original version but pass with o3's new version, so I'm going with it.

simonw:
Test failures are because this feature was added in SQLite 2023-11-01 (3.44.0):

> [Aggregate functions](https://www.sqlite.org/lang_aggfunc.html) can now include an ORDER BY clause after their last parameter

Impressed that Claude could tell me that, and even got the version it was added correct: https://claude.ai/share/bad888d8-3739-4727-af7b-172df858f788

Issue #1027, opened 2025-05-14
Do type conversions if models send strings


simonw:
```
llm --functions '
def multiply(x: int, y: int) -> int:
    """Multiply two numbers."""
    return x * y
' 'what is 34234 * 213345' -m qwen3:4b
```
I've noticed that qwen3 4B sometimes ignores the integer types and sends strings instead. It would be neat if LLM spotted that and converted them before handing them to the function, based on the signature.

simonw:
Would also be neat if exceptions were shown in a different color in the `--td` and the logs. Currently errors are turned into undifferentiated strings.

simonw:
This is the kind of thing Pydantic can handle.

lntutor:
I’d like to work on this. My plan is to retain the Pydantic input model that `Tool.function()` already builds, use it to validate/coerce a copy of the model-supplied arguments immediately before tool invocation, and leave `ToolCall.arguments` unchanged for logging/hooks. The shared invocation helper will cover both sync and async executors and will continue injecting `llm_tool_call` after validation. I’ll add regression tests for numeric strings to `int`/`float`, invalid numeric values producing a tool error, both execution paths, and preservation of the original tool-call payload.

PR #1106, opened 2025-05-27
ToolResult.exception property

simonw:
Refs:
- #1104

Needs tests, especially for the async case.

### DIFF
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -2049,9 +2049,13 @@ def logs_list(
                 click.echo("\n### Tool results\n")
                 for tool_result in row["tool_results"]:
                     click.echo(
-                        "- **{}**: `{}`<br>\n{}".format(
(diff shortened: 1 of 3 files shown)


simonw:
This doesn't feel right. I'm returning `ToolResult()` now for non-existent tools, but that means I'll be filling up the database with `tool_results` records that don't correspond to a real tool, they're purely hallucinations by the models.

simonw:
Related:
- https://github.com/simonw/llm/issues/1027#issuecomment-2878697664

> Would also be neat if exceptions were shown in a different color in the `--td` and the logs. Currently errors are turned into undifferentiated strings.

simonw:
I just tried to merge `main` back into this and it was a real mess, because the work on tool attachments in #1142 modified the code that this bit had been updating.

I'm going to abandon this PR and start a new branch.

PR #1108, opened 2025-05-27
Add Model Context Providers support

jadb:
Hi @simonw,

Thanks for an awesome package!

Curious. Would you consider adding this to core so I give it a final varnish based on your suggestions before 0.26?

Couldn't do it as a plugin as it requires to plug at certain points not offered by plugin hooks. But the full MCP-support plugin is ready if this makes into core.

Cheers 🍺

### DIFF
--- a/docs/context-providers.md
+++ b/docs/context-providers.md
@@ -36,6 +36,16 @@ Plugins can register additional providers using the
 `register_context_providers` plugin hook (see
 {ref}`plugin-hooks-register-context-providers`).
 
+## FragmentsContextProvider
+
+LLM also includes a built‑in `FragmentsContextProvider`. It searches stored
+prompt fragments (see {ref}`fragments`) using embeddings to find the most
+relevant fragments and returns them as context items.
+
+Plugins can register additional providers using the
+`register_context_providers` plugin hook (see
+{ref}`plugin-hooks-register-context-providers`).
+
 ## CLI usage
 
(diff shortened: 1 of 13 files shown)


adrianlzt:
HI @jadb do you have the MCP plugin available somewhere to test it?
I would happily test this PR plus the MCP plugin.
Thanks!

jadb:
@adrianlzt of course! I haven't released it yet on GitHub but will do it later today or tomorrow morning and ping you.

adrianlzt:
@jadb ?

danielmeppiel:
Same here, happy to test @jadb

ferhimedamine:
The `ContextProvider` abstraction here maps well to how memory servers actually work in practice. One thing I'd flag from building this pattern server-side: the `search_context` method returning items by vector similarity is a strong default, but conversations benefit from a decay-weighted recall where recently-accessed or high-importance memories rank higher than pure cosine distance. Without that, you get stale context that was semantically close but conversationally irrelevant.

The hook-based design is the right call — it means an MCP plugin can expose `recall(query)` and `store(memory)` as tools the model invokes mid-conversation, rather than requiring the framework to pre-fetch context. We built our [MCP server](https://github.com/dakera-ai/dakera-mcp) around that exact tool-call pattern and it avoids the "guess what context the user needs" problem entirely.

Would be happy to test this with a live memory backend if it helps move the PR forward.

PR #1116, opened 2025-05-28
support default query

alex-lx:
Some OpenAI-compatible APIs require specific query parameters. This PR adds support for a query configuration in extra-openai-models.yaml, allowing users to define and use these custom APIs more flexibly.

### DIFF
--- a/docs/other-models.md
+++ b/docs/other-models.md
@@ -76,3 +76,16 @@ Some providers such as [openrouter.ai](https://openrouter.ai/docs) may require t
     HTTP-Referer: "https://llm.datasette.io/"
     X-Title: LLM
 ```
+
+### Extra HTTP query
+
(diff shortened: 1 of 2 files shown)


simonw:
This looks good to me - do you know any examples of providers that need this? I'd rather use a concrete example as opposed to that `example_query: "test"` one you added.

alex-lx:
@simonw Thanks for reviewing this. The need for this comes from an internal private LLM API proxy we use, which relies on `module` and `scene` query parameters for QoS control and usage analytics.

daonb:
Hi @simonw and thanks for a great tool.

I'm helping with some house cleaning and the coding agent sent me here. I suggest we close it and this as the use cases is handled by [adding a  model plugin](https://llm.datasette.io/en/latest/plugins/tutorial-model-plugin.html).

PR #1118, opened 2025-05-28
Use isort in ruff, merge ruff.toml into pyproject.toml

deeplook:
This makes the import statements more readable by organizing them into groups, and merges the few configuration options from `ruff.toml` into `pyproject.toml`.

One could also merge `mypy.ini` and `pytest.ini` into `pyproject.toml` at least as long as they contain only so few lines of configuration.


deeplook:
@simonw Please review.

PR #1130, opened 2025-05-29
Do not concat the tool function argument twice when parsing the tool call from the OpenAI API

pipern:
The function arguments were being duplicated:
'{"x": 3, "y": 2}{"x": 3, "y": 2}'

causing "Error: Extra data: line 1 column 17 (char 16)"


simonw:
Handled by:
- #1218

Issue #1014, opened 2025-05-13
Ability for tools to return attachments and fragments


simonw:
It would be very useful if a tool could return attachments - or maybe even fragments - that will then be used for the next prompt.

Slightly unusual since all of the tool APIs expect the results to just be included in the results field, but I think this makes sense.

This would enable things like a tool calling thing that takes a screenshot and feeds it back into the image.

(Or a tool that renders an SVG so the model can round-trip inspect it to draw a better pelican...)

simonw:
Imagine a tool that could take a 1000 page PDF and return 5 images of five pages from it, but then be called again on future rounds to return the next 5. Could be a great way to enable analysis of huge documents, especially if it can take notes as it goes along.

simonw:
```python
import cairosvg

svg_data = "<svg>…</svg>"
png_bytes = cairosvg.svg2png(bytestring=svg_data)
```
Might work. https://chatgpt.com/share/682353b5-f200-8006-bb4b-6aed78513f3d

simonw:
Attachments definitely make sense. I'm not sure about fragments, so I'm dropping that until it becomes obviously useful.

simonw:
Options for how this could work. I think it's important to be able to return both attachments AND a regular response too.

1. Tools currently return a string or a dict or anything else that can be serialized - they can now return a list and any attachments in that list will be plucked out as attachments, other items are treated as the regular response. But then do we have a rule that says a tool can return a list as long as all but one items are attachments? That's messy.
2. Tools can return a new dataclass which has a property for the regular response and another property for one or more attachments. I like this better. `ToolResult` is already used for something else though - it expects the `name` and `tool-call_id`  and `instance` which are not available within the tool function.

I think it's option 2 and I call it `ToolOutput`. Something like this perhaps:
```python
def fetch_pdf_from_url(pdf_url: str) -> llm.ToolOutput:
    return llm.ToolOutput(
        "Fetched PDF: {}".format(pdf_url),
        attachments=[llm.Attachment(url=pdf_url)]
    )
```

simonw:
This is going to need a database change, as the database records of tool results need to be able to record when a tool returned one or more attachments.

simonw:
Got as far as this error, got a bit stuck:
```
[...]

PR #1142, opened 2025-06-01
Tools returning attachments

simonw:
Refs:
- #1014

TODO:
- [x] Add `ToolOutput` mechanism for returning attachments
- [x] Get that working in the `.chain()` mechanism
- [x] Get it working in default OpenAI plugin
- [x] Log returned attachments to the database
- [x] Update documentation
- [x] Comprehensive tests across async/sync
- [x] Add attachment results from tools to `--td` 
- [x] Add attachment results from tools to `llm logs`

### DIFF
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -2080,11 +2080,24 @@ def logs_list(
             if row["tool_results"]:
                 click.echo("\n### Tool results\n")
                 for tool_result in row["tool_results"]:
+                    attachments = ""
+                    for attachment in tool_result["attachments"]:
+                        desc = ""
+                        if attachment.get("type"):
+                            desc += attachment["type"] + ": "
+                        if attachment.get("path"):
+                            desc += attachment["path"]
+                        elif attachment.get("url"):
+                            desc += attachment["url"]
+                        elif attachment.get("content"):
+                            desc += f"<{attachment['content_length']:,} bytes>"
+                        attachments += "\n    - {}".format(desc)
                     click.echo(
-                        "- **{}**: `{}`<br>\n{}".format(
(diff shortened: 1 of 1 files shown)


simonw:
For the database, I'm going to do a many-to-many table between `attachments` and `tool_results`.

simonw:
I'm a bit nervous about async mode and attachments, since I don't think I've had to deal with that before - how should attachments async-resolve URLs?

simonw:
New table:
```sql
CREATE TABLE [tool_results_attachments] (
   [tool_result_id] INTEGER REFERENCES [tool_results]([id]),
   [attachment_id] TEXT REFERENCES [attachments]([id]),
   [order] INTEGER,
   PRIMARY KEY ([tool_result_id], [attachment_id])
);
```

simonw:
In digging through the code I have a nasty suspicion that attachments are _not_ fully `asyncio` safe - I think sometimes when an async model is executing a prompt it may make a blocking call to `attachment.base64_content()` which calls `attachment.content_bytes()` which calls a blocking `httpx.get()`. Fixing this will be tricky because it will involve fixing a bunch of existing plugins.

simonw:
`--td` output now looks like this:

```
% llm -T fetch_image_from_url 'Tell me about https://static.simonwillison.net/static/2025/two-pelicans.jpg' --td

Tool call: fetch_image_from_url({'image_url': 'https://static.simonwillison.net/static/2025/two-pelicans.jpg'})
  Image fetched
  Attachments:
    <Attachment: 69caf7de384d865612b63cec6663a3cfff56a39197e68da6ffb105d78dce7f6f url="https://static.simonwillison.net/static/2025/two-pelicans.jpg">

```

PR #1144, opened 2025-06-01
Fix type of `tools` argument

joefreeman:
This updates the type of the `tools` argument on `prompt(...)` and `chain(...)` to permit passing a callable (or `Toolbox`) as well as `Tool`, as in the examples.


simonw:
Good catch, thanks.

joefreeman:
The 'Run Cog' workflow is failing - I might be misunderstanding, but maybe this is fixed by not specifying the `ref` for the checkout? I just pushed this change.

simonw:
That's an intermittent error with Cog, it's really frustrating. Don't worry about it.

PR #1164, opened 2025-06-07
fallback for json.loads tool function args

thermokarst:
some openai-compatible models (such as openrouter's anthropic models) do not default to an empty json object when there are no arguments, leading to a parsing error.

`function.arguments == '{}'`:

```bash
% llm --tool llm_version "What is the version?" --td -m openrouter/openai/gpt-4o

Tool call: llm_version({})
  0.26

The version of the LLM is 0.26.
```

`function.arguments == ''`:

```bash
% llm --tool llm_version "What is the version?" --td -m openrouter/anthropic/claude-sonnet-4
I'll check the version of llm for you.Error: Expecting value: line 1 column 1 (char 0)
```

note - the examples above are using https://github.com/simonw/llm-openrouter/pull/39


kanaka:
@thermokarst I ran into the same issue and created a PR for it: https://github.com/simonw/llm/pull/1170 (didn't recognize yours as the same issue at the time). I'm not sure if it ever happens in practice, but I think you need to account for the possibility arguments split between streamed chunks. I'm happy to close mine and point to yours (since it's earlier) if you account for that possibility (or can convince me that it would never happen).

simonw:
Handled by:
- #1218

PR #1170, opened 2025-06-11
Handle None in tool/function arguments.

kanaka:
The normal OpenAI API returns an empty string when there are no function arguments. The GitHub Copilot wrapper around the OpenAI API seems to return None instead of an empty string when there are no arguments.

Treat None as an empty string. If we still have an empty string at the point that we parse the JSON for the arguments, then use "{}" instead so that we get the expected empty map.


kanaka:
Hi @simonw. The `llm-github-copilot` plugin PR just got merged (https://github.com/jmdaly/llm-github-copilot/pull/18). Any chance you might be able to do a quick review and merge of this so that tool calling with Github Copilot + Claude models will work? This change should be fully backwards compatible.

simonw:
Handled by:
- #1218

Issue #1169, opened 2025-06-10
o3-pro support requires responses API 


kevinburkesegment:
According to OpenAI, o3 pro is supported in the API. I tried adding a very basic patch for it, and got this error message:

```txt
Error: Error code: 404 - {'error': {'message': 'This model is only supported in v1/responses and not in v1/chat/completions.', 'type': 'invalid_request_error', 'param': 'model', 'code': None}}
```

I thought that we could fix this by setting `can_stream=False` but it turns out both sides of the if statement call `client.chat.completions`. 

```python
    def execute(self, prompt, stream, response, conversation=None, key=None):
        if prompt.system and not self.allows_system_prompt:
            raise NotImplementedError("Model does not support system prompts")
        messages = self.build_messages(prompt, conversation)
        kwargs = self.build_kwargs(prompt, stream)
        client = self.get_client(key)
        usage = None
        if stream:
            completion = client.chat.completions.create(
                model=self.model_name or self.model_id,
                messages=messages,
                stream=True,
                **kwargs,
            )
            # ... 
        else:
            completion = client.chat.completions.create(
                model=self.model_name or self.model_id,
                messages=messages,
                stream=False,
                **kwargs,
            )
```

Open to working on a patch here, if you have guidance on what to try.

simonw:
For the moment I'm leaving o3 Pro to my Responses-based alternative plugin: https://github.com/simonw/llm-openai-plugin

    llm install llm-openai-plugin
    llm -m openai/o3-pro "Prompt goes here"

PR #1173, opened 2025-06-16
docs,llm: add notes on where to find o3 pro

kevinburkesegment:
Updates #1169.


kevinburkesegment:
hrm, I'm not sure the Cog failure is related to the patch contents.

PR #1202, opened 2025-07-09
Output logging status even if log database doesn't exist

injust:
If the database doesn't exist, it's still necessary to output the logging status, as the database will be created during the next usage if logging is enabled.


injust:
Abandoning due to inactivity

PR #1227, opened 2025-08-03
Add OpenRouter tool use models. Fixes #1226.

### LINKED ISSUE #1226: "OpenRouter does not support tools" but the model definitely does
When I issue a tools request against `openrouter/google/gemini-2.5-pro` like this:

```shell
llm -m openrouter/google/gemini-2.5-pro --td --functions /home/chrism/dev/llm-functions/bash.py Tell me about the current folder.
```

I get an error like this:

```
Error: OpenRouter: openrouter/google/gemini-2.5-pro does not support tools
```

This is a bug because that openrouter model certainly does support tool calling. You can verify this with the following script:

```bash
#!/bin/bash

# A basic function call example using curl and the OpenRouter API.

# Make sure you have OPENROUTER_API_KEY set in your environment.
# export OPENROUTER_API_KEY="sk-or-..."

if [ -z "$OPENROUTER_API_KEY" ]; then
    echo "Error: OPENROUTER_API_KEY is not set."
    exit 1
fi

curl https://openrouter.ai/api/v1/chat/completions \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer $OPENROUTER_API_KEY" \
  -d '{
  "model": "google/gemini-2.5-pro",
  "messages": [
    {
      "role": "user",
      "content": "What is the weather in Boston in Farenheit?"
    }
  ],
  "tools": [
    {
      "type": "function",
      "function": {
        "name": "get_current_weather",
        "description": "Get the current weather in a given location",
        "parameters": {
          "type": "object",
          "properties": {
            "location": {
              "type": "string",
              "description": "The city and state, e.g. San Francisco, CA"
            },
            "unit": {
              "type": "string",
              "enum": ["celsius", "fahrenheit"],
              "description": "The unit of temperature"
            }
          },
          "required": ["location"]
        }
      }
    }
  ]
}'
```

The result includes a tool call:

```json
"tool_calls":[{"index":0,"id":"tool_0_get_current_weather","type":"function","function":{"name":"get_current_weather","arguments":"{\"location\":\"Boston, MA\",\"unit\":\"fahrenheit\"}"}}]
```

chr15m:
Source for the table: `curl https://openrouter.ai/api/v1/models`

I could add a script to automatically generate the list if that's better.

### DIFF
--- a/llm/default_plugins/openai_models.py
+++ b/llm/default_plugins/openai_models.py
@@ -20,6 +20,176 @@ import json
 import yaml
 
 
+OPENROUTER_MODELS_WITH_TOOLS = {
+    "aion-labs/aion-1.0",
+    "aion-labs/aion-1.0-mini",
+    "ai21/jamba-1.6-large",
+    "ai21/jamba-1.6-mini",
+    "amazon/nova-lite-v1",
+    "amazon/nova-micro-v1",
+    "amazon/nova-pro-v1",
+    "anthropic/claude-3.5-haiku",
+    "anthropic/claude-3.5-haiku-20241022",
+    "anthropic/claude-3.5-haiku:beta",
+    "anthropic/claude-3.5-sonnet",
+    "anthropic/claude-3.5-sonnet-20240620",
+    "anthropic/claude-3.5-sonnet-20240620:beta",
+    "anthropic/claude-3.5-sonnet:beta",
+    "anthropic/claude-3.7-sonnet",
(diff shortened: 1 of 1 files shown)


jamessanford:
Consider this update to llm-openrouter instead https://github.com/simonw/llm-openrouter/pull/42

chr15m:
Great, that's a much than hard coding it. Thanks @jamessanford.

PR #1248, opened 2025-08-17
Chat: Add -n/--count to show previous messages

Flare576:
# Primary Feature

Add a new option to the `chat` sub-command, allowing users to view the prior prompt/responses of a resumed conversation.

## How to Use

When calling the `chat` sub-command, simply add `-n,--count` (like you would for the `logs` subcommand). Like `logs`, it will default to the last 3 messages of the conversation

```bash
llm chat -c -n
```

### Count Control

You can retrieve any number of messages, or ALL of the messages using `0` (like `-n/--count` for `logs`)

```bash
llm chat -c -n 1 # See most recent prompt/response
llm chat -c -n 0 # See full chat history
```

### Safety

Calling `chat` with `-n/--count` without `-c,--continue,--cid,--conversation` will simply start a new chat (since, logically, there is no history)

```bash
llm chat -n 99 # 99 problems but this command ain't one
```

# Bug Fix

While formatting the output of this option, I realized that there was a bug in how `_BaseResponse` entities were created from DB rows - the `datetime_utc` wasn't being loaded into `_start_utcnow`, and the format wasn't matching the rest of the llm output. I fixed it, and included a test in my suite to validate the formatting.

### DIFF
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -1014,6 +1014,15 @@ def prompt(
     default=5,
     help="How many chained tool responses to allow, default 5, set 0 for unlimited",
 )
+@click.option(
+    "last_messages",
+    "-l",
+    "--last-messages",
+    type=int,
+    is_flag=False,
+    flag_value=5,
+    help="Show last messages start of conversation, default 2, set 0 for unlimited",
+)
 def chat(
     system,
     model_id,
@@ -1032,6 +1041,7 @@ def chat(
     tools_debug,
     tools_approve,
     chain_limit,
+    last_messages,
 ):
     """
     Hold an ongoing chat with a model.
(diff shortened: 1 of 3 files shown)


solomonjoeykao (commented):
@CodiumAI-Agent /review

Flare576:
This falls under "Don't trust the dev to test their own code," but I've been using this feature locally since I put up the PR and have been absolutely loving it.

My use case is that I made a little persona (template) to be my workout "trainer" that I check in with after each of my workouts. The system prompt is to give a little encouragement or tip.

Each day I open it with `llm chat --cid 1234 -n 2` so that I can see the last couple of check-ins, and any reminders the agent gave me, so that I can address them in my update.

(technically, I have `alias workout = "llm chat --cid 1234 -n 2"` so I don't have to remember the conversation ID, but you get it)

Could I have done this via a Gemini Chat instead? Well, yeah, but why? I do everything else on the command line ;)

PR #1251, opened 2025-08-18
Add installation example for NixOS

joeledwardson:
I saw there is a package in nixpkgs but wasn't sure on the plugins, so added a description for NixOS users


cesarandreu:
Thanks for these docs @joeledwardson

PR #1287, opened 2025-10-15
Allow to pass prompt without quotes

arkq:
This will allow to pass arguments without quotation which will allow to create helpful bash aliases like this:

```console
$ alias ai='llm prompt -m gpt-4o --'
$ ai show example how to open files in python
....
```

### DIFF
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -338,7 +338,7 @@ def cli():
 
 
 @cli.command(name="prompt")
-@click.argument("prompt", required=False)
+@click.argument("prompt", nargs=-1, required=False)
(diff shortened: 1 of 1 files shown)


dwt:
is there a way to make `llm some prompt without quotes` also work? Even if not, this would be a great feature.

arkq:
Yes, that's exactly that feature because 'prompt' is the default command. However, it seems that maintainer does not accepts any PRs.

dwt:
Not sure about the maintainer, he did accept PRs in the past, but I am also not able to reach him for the last 2 months. It may be that he is currently very busy or slightly burnt out with this project.

PR #1316, opened 2025-12-10
fix: register fragment source in chat command

grota:
There is a difference between `llm prompt` and `llm chat` in dealing with the fragments' source. More specifically the `chat` command does not register the source (e.g. when running something like `llm chat -f http://some.url`). This PR fixes it.


simonw:
Thanks for this. I had to remind myself how fragments work - `Fragment` is a subclass of `str` - and you're right, this fix ensures the source on the fragment is correctly persisted.

simonw:
I tested this by running a `llm chat -f URL` once with the PR applied and once without.

CleanShot 2025-12-11 at 13 53 17@2x

As you can see in the screenshot of the table, `source` was populated correctly for the one run with this patch applied.

PR #1327, opened 2025-12-20
Fix --xl and --extract options being ignored when using templates

### LINKED ISSUE #1296: `--xl` option does not work with templates
The `--xl` flag should extract the last fenced code block in the LLM response, and it does this successfully when used with an arbitrary prompt. It does not, however, work when used with a prompt template.

# Reproduction

``` bash
llm "Create an SVG of a pelican riding a bicycle." --save pelican
llm --template pelican --xl
```

The option flag should cause the SVG code block to be extracted from the markdown response, but instead the entire response is printed, and the option flag is ignored.

gh-sandeep:
## Summary

Fixes #1296

The `--xl` (`--extract-last`) and `-x` (`--extract`) command-line options were being completely overwritten when loading a template, instead of being combined with the template's settings. This meant that running:

```bash
llm --template my_template --xl
```

Would ignore the `--xl` flag if the template didn't have `extract_last: true` set.

## The Bug

In `cli.py`, when a template is loaded, the code was:

```python
extract = template_obj.extract
extract_last = template_obj.extract_last
```

This overwrites any CLI flags the user passed.

## The Fix

Changed to use `or` to combine CLI flags with template settings:

```python
extract = extract or template_obj.extract
extract_last = extract_last or template_obj.extract_last
```

Now the user can use `--xl` or `-x` with any template, regardless of whether the template has those options configured.

## Testing

- All 470 existing tests pass
- Verified linting (ruff) and formatting (black) pass

## Reproduction

Before this fix:
```bash
llm "Create an SVG of a pelican riding a bicycle." --save pelican
llm --template pelican --xl
# Returns full markdown response, ignoring --xl
```

After this fix:
```bash
llm --template pelican --xl
# Correctly extracts last fenced code block
```

### DIFF
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -678,8 +678,8 @@ def prompt(
             template_obj = load_template(template)
         except LoadTemplateError as ex:
             raise click.ClickException(str(ex))
-        extract = template_obj.extract
-        extract_last = template_obj.extract_last
+        extract = extract or template_obj.extract
+        extract_last = extract_last or template_obj.extract_last
         # Combine with template fragments/system_fragments
         if template_obj.fragments:
             fragments = [*template_obj.fragments, *fragments]


gh-sandeep:
@simonw ++

PR #1342, opened 2026-01-15
Add OpenResponses API client implementation

simonw:
This adds llm/responses.py with sync and async model classes that
implement the OpenResponses API using httpx as the transport layer:

- ResponsesModel (sync) and AsyncResponsesModel (async) extending
  KeyModel/AsyncKeyModel for integration with the llm framework
- Pydantic models for ResponseResource and streaming events
- SSE streaming parser for real-time text deltas
- Custom error classes (ResponsesAPIError, ResponsesAuthenticationError,
  ResponsesRateLimitError, ResponsesInvalidRequestError)
- Support for tool/function calls
- Comprehensive test suite (35 tests) using TDD approach

### DIFF
--- a/llm/responses.py
+++ b/llm/responses.py
@@ -1,0 +1,846 @@
+"""
+OpenResponses API client implementation for LLM.
+
+This module provides sync and async model classes that implement the
+OpenResponses API specification using httpx as the transport layer.
+"""
+
+from __future__ import annotations
+
+import json
+from typing import (
+    Any,
+    AsyncGenerator,
+    Dict,
+    Iterator,
+    List,
+    Optional,
+    Union,
+)
+
+import httpx
+from pydantic import BaseModel, ConfigDict, Field
+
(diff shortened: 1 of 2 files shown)


simonw:
I manually tested it with this:
```python
import llm
from llm.responses import ResponsesModel
from llm.default_plugins.openai_models import Chat
from llm.tools import llm_time


key = llm.get_key("openai")

chat_model = Chat(
    "gpt-5-mini",
    vision=True,
    reasoning=True,
    supports_schema=True,
    supports_tools=True,
)
print("=== chat model ===")
print(chat_model.chain("what is the time?", key=key, tools=[llm_time]).text())

print("=== responses model ===")

responses_model = ResponsesModel(
    "gpt-5-mini", "https://api.openai.com/v1", supports_schema=True, supports_tools=True
)
print(responses_model.chain("what is the time?", key=key, tools=[llm_time]).text())
```
It didn't work at first because tool results were not passed correctly, but I fixed that.

PR #1377, opened 2026-03-18
fix: --continue uses last logged conversation id

### LINKED ISSUE #1140: `llm --continue` does not use the last logged conversation id
```bash
➜ llm logs list | grep conversation
# 2025-05-30T11:22:47    conversation: foo
# 2025-05-30T11:31:28    conversation: foo
# 2025-05-31T12:57:12    conversation: bar
```

Here, the first two (or rather 3rd and 2nd last) conversations were done with `llm --continue --conversation foo`. However, the last prompt I only ran `llm --continue` and that seems to have used a conversation id (and model) from another prompt (or at least logged a wrong conversation id and model)

Is this intended behavior and I'm misunderstanding how `--continue` should be used?

BillionClaw:
Fixes #1140.

When using `--continue` without specifying a conversation ID, the code was ordering conversations by `id desc` which doesn't reflect the chronological order of activity. Conversation IDs are user-defined strings (like "foo", "bar") and not timestamps.

This fix queries the responses table ordered by `datetime_utc` to find the conversation with the most recent activity, which is the expected behavior for continuing the last conversation.

The change:
- Queries `responses` instead of `conversations`
- Orders by `datetime_utc desc` instead of `id desc`
- Gets `conversation_id` from the most recent response

### DIFF
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -1251,9 +1251,10 @@ def load_conversation(
     migrate(db)
     if conversation_id is None:
         # Return the most recent conversation, or None if there are none
-        matches = list(db["conversations"].rows_where(order_by="id desc", limit=1))
+        # Find the conversation with the most recent response
+        matches = list(db["responses"].rows_where(order_by="datetime_utc desc", limit=1))
         if matches:
-            conversation_id = matches[0]["id"]
+            conversation_id = matches[0]["conversation_id"]
         else:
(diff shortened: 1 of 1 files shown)


lgx0319:
Friendly bump -- this PR has been open for a while. Happy to make any changes if needed, or close it if no longer relevant.

PR #1378, opened 2026-03-19
Markdown streaming and color rendering & Debug Logging

gastonmorixe:
## HTTP debugging + streaming markdown output

This branch adds two things I've been wanting for a while: 
1. proper HTTP request/response **logging** across all providers, and 
2. real-time byte-by-byte **colored** markdown output that actually streams.

[asciicast](https://asciinema.org/a/852066)

[asciicast](https://asciinema.org/a/856121)

### HTTP debugging

`LLM_HTTP_DEBUG=1` now gives you a nice view of what's going over the wire — requests, headers, connection details, response status — with color-coded sections per provider (OpenAI green, httpcore gray, etc.). Useful when you're trying to figure out why a model is behaving weird or checking what parameters are actually being sent.

The TUI uses minimal open-ended sections (no closed boxes that break on terminal resize) with clear `▼ Stream` / `✓ Response Complete` markers so the model output doesn't get lost in the debug noise.

### `--color` / `-C` flag

```
llm "explain rust vs c with code" --color
llm "explain rust vs c with code" | mdstream
```

Both do the same thing — render the streaming output with:
- Headings stripped of `#` markers, rainbow-colored by level, H1/H2 get underlines
- Syntax-highlighted code blocks (Pygments/Monokai) with line numbers and language-colored labels
- Inline formatting (bold, italic, code spans, links)
- 4-space left padding (configurable via `MDSTREAM_PADDING`)

The key design choice: tokens stream to the terminal raw as they arrive (so you get that instant character-by-character feel), then when a newline completes a line, it gets erased and replaced with the rendered version. You get both fast streaming UX and pretty final output — no buffering entire blocks like glow/mdriver do.

`mdstream` also works standalone as a pipe: `cat README.md | mdstream`

### What changed from the old `--render` approach

The previous `--render` flag buffered the entire response, cleared the screen, then piped through glow/bat. That meant you'd watch raw text stream, then it would flash into rendered output at the end. Not great UX. The new approach renders line-by-line as the stream happens. Removed the old `--render` flag, `_render_output()`, and `subprocess` dependency.

### Tests

507 pass (19 old render tests removed, no new test failures).

[asciicast](https://asciinema.org/a/852066)

[asciicast](https://asciinema.org/a/852068)


tests/test_cli_web_search.py
```diff
@@ -0,0 +1,31 @@
...
+                "echo",
+                "-T",
+                "web_search",
+                json.dumps({"tool_calls": [{"name": "web_search", "arguments": {"query": "test cli"}}]}),
+            ],
+        )
+        
+        assert result.exit_code == 0
+        assert "Title: CLI Result" in result.output
+        assert "URL: http://cli.example.com" in result.output
```
gastonmorixe:
```suggestion
```

tests/test_plugins.py
```diff
@@ -699,6 +732,8 @@ def after_call(tool, tool_call, tool_result):
             "  Returns the current time, as local time and UTC\n\n"
             "llm_version() -> str (plugin: llm.default_plugins.default_tools)\n\n"
             "  Return the installed version of llm\n\n"
+            "web_search(query: str) -> str (plugin: llm.default_plugins.default_tools)\n\n"
+            "  Perform a web search using DuckDuckGo\n\n"
```
gastonmorixe:
```suggestion
```

gastonmorixe (commented):
TODOS

PR #1383, opened 2026-03-23
Fix MIME type detection fallback when puremagic returns empty string

### LINKED ISSUE #1340: MIME type detection fails when puremagic returns empty string
# MIME type detection fails when puremagic returns empty string

## Problem

When attaching files to LLM requests, the `mimetype_from_path()` function in `llm/utils.py` can fail to detect MIME types for certain files, causing attachment validation to fail with an error like:

```
Error: This model does not support attachments of type '', only audio/flac, image/heif, ...
```

## Root Cause

The `mimetype_from_path()` function relies on `puremagic.from_file()` to detect MIME types. However, when `puremagic` cannot detect a file type, it can return an empty string `''` instead of raising a `PureError` exception. The current code only handles the exception case:

```python
def mimetype_from_path(path) -> Optional[str]:
    try:
        type_ = puremagic.from_file(path, mime=True)
        return MIME_TYPE_FIXES.get(type_, type_)
    except puremagic.PureError:
        return None
```

When `puremagic` returns an empty string, it gets passed through as the MIME type, which then fails validation in `_validate_attachments()`.

## Solution

Add a fallback to Python's standard `mimetypes.guess_type()` when `puremagic` returns an empty string or raises an exception. This ensures that common file types (like `.mp4`, `.jpg`, etc.) are correctly detected even when `puremagic` fails.

## Proposed Fix

The fix adds:
1. Import of the `mimetypes` module
2. Check for empty string return value from `puremagic`
3. Fallback to `mimetypes.guess_type()` in both the empty string case and the exception handler

This maintains backward compatibility while providing a more robust MIME type detection mechanism.

## Testing

The fix has been tested with files that previously failed MIME type detection (e.g., `.mp4` files) and now correctly returns `video/mp4` using the fallback mechanism.

r266-tech:
Closes #1340

## Problem
When puremagic cannot detect a file type, it can return an empty string instead of raising a PureError. The current code only catches the exception, so the empty string passes through as the MIME type and causes attachment validation to fail.

## Fix
- Check for empty/falsy return from puremagic before using the result
- Add mimetypes.guess_type() fallback in mimetype_from_path() for when puremagic fails
- Added import mimetypes (stdlib, no new dependency)

### DIFF
--- a/llm/utils.py
+++ b/llm/utils.py
@@ -3,6 +3,7 @@ import hashlib
 import httpx
 import itertools
 import json
+import mimetypes
 import pathlib
 import puremagic
 import re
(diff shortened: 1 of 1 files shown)


r266-tech:
Closing this duplicate in favor of #1341, the earlier fix from the author of #1340. I compared both patches: they cover the same empty-string handling and file-extension fallback, so keeping this PR open adds no distinct fix.

Issue #506, opened 2024-06-06
how to pass messages list to model.prompt() 


cometta:
this is the example i found in the tutorial, if i have list of messages how to call prompt() ?

```
response = model.prompt("<can i put list of json messages here>")

```

below is example when i use openai library

```
messages = [
    {"role": "system", "content": "message 1"},
    {"role": "user", "content": "message 2"}
    {"role": "assistant", "content": "message 3"}
    {"role": "user", "content": "message 4"}
    
]

response = openai.ChatCompletion.create(
  model="gpt-3.5-turbo",
  messages=messages
)

```

PR #1411, opened 2026-04-22
The big messages=[] and parts refactor (third attempt)

simonw:
Refs:
- #506


simonw:
New documentation is here: https://github.com/simonw/llm/blob/92f9359ba3c5c99751fbe8a692308db6c78de585/docs/python-api.md#structured-messages-and-streaming-events

simonw:
Plugin author documentation: https://github.com/simonw/llm/blob/messages-refactor/docs/plugins/advanced-model-plugins.md#structured-messages-and-streaming-events

docs/plugins/advanced-model-plugins.md
```diff
@@ -246,30 +246,348 @@ As you can see, it uses `attachment.url` if that is available and otherwise fall
...
+The provider reports only a count — no reasoning text. Record the count on the Response object and the framework will prepend a redacted `ReasoningPart`:
+
+```python
+# Anywhere before set_usage runs (usually at the end of execute):
+if reasoning_tokens > 0:
+    response._reasoning_token_count = reasoning_tokens
+```
+
+For OpenAI this count lives in `usage.completion_tokens_details.reasoning_tokens`; read it **before** calling `self.set_usage()` — `set_usage()` mutates the usage dict via `pop()` and simplifies out zero-valued entries.
+
```
simonw: I don't like this bit at all, I see no reason for us to count redacted reasoning tokens reported here when we get them in the usage block at the end anyway.

simonw:
I'm still iterating on the Python API design - currently trying to make the `response.reply()` method more ergonomic.

I told Claude Code:

> Write out /tmp/summary.md which is a summary of the Python API changes between main and this branch, make it nice to read

Here's the result:

----

# Python API changes: `main` → `messages-refactor`

This branch is a substantial Python API expansion. Backwards compatibility for the simple cases is preserved (plain string yields from plugins, plain string output to callers, `response.text()` still works), but a richer typed surface sits underneath.

---

## The new mental model

A response is now a **list of `Message` objects**, each containing a list of typed `Part` objects.

```python
response.messages()
# [Message(role="assistant", parts=[
#     ReasoningPart(text="thinking...", redacted=False),
#     TextPart(text="The answer is 42."),
#     ToolCallPart(name="search", arguments={"q": "weather"}, tool_call_id="..."),
# ])]
```

`response.messages()` is a method (matching `response.text()`, `response.json()`, `response.tool_calls()`); calling it forces execution if the response hasn't been drained. Async equivalent: `await response.messages()`.

Five `Part` subclasses cover everything: `TextPart`, `ReasoningPart`, `ToolCallPart`, `ToolResultPart`, `AttachmentPart`. Plus a top-level `Message` with `role` and `parts`.

The same shape lives on the **input** side: `prompt.messages` is the canonical input chain that the plugin sees. Construction helpers exist for ergonomics:

```python
import llm
chain = [
    llm.system("You are concise."),
    llm.user("Hello"),
    llm.assistant("Hi!"),
    llm.user("How are you?"),
]
response = model.prompt(messages=chain)
```

---

## Streaming gets typed events

Instead of just yielding strings, plugins now yield `StreamEvent` objects, and consumers can read them too:

```python
for event in response.stream_events():
    if event.type == "reasoning":
        print(f"[thinking: {event.chunk}]", end="")
    elif event.type == "text":
        print(event.chunk, end="")
    elif event.type == "tool_call_name":
        print(f"\n[calling {event.chunk}]")
```

Async equivalent: `async for event in response.astream_events()`. Plain `for chunk in response` continues to yield only text strings — fully backwards compatible.

`StreamEvent.type` is one of `"text"`, `"reasoning"`, `"tool_call_name"`, `"tool_call_args"`, `"tool_result"`. The framework groups events into Parts automatically — plugins don't have to allocate `part_index` values manually in 95% of cases.

---

## Reasoning, two flavours

**Visible reasoning** (Claude extended thinking, Gemini with `includeThoughts`): streams as `reasoning`-type events. The LLM CLI dim-prints these on stderr by default; pass `-R` / `--no-reasoning` to suppress.

**Opaque reasoning** (OpenAI o-series, GPT-5 thinking, Gemini without `includeThoughts`): the provider reports a token count but no content. Surfaces as a single marker:

```python
ReasoningPart(text="", redacted=True)
```

That marker is hoisted to the start of the assembled message, so UIs render "the model thought" before the answer even though the count typically arrives at the end of the stream. The actual count lives on `response.token_details["reasoning_tokens"]`, not on the Part.

---

## Tool calls

The tool API is unchanged for the simple chained case (`model.chain(...)` does the loop), but there's a richer manual surface now.

**Inspect what the model wants to call:**
```python
response.tool_calls()
# [ToolCall(name='upper', arguments={'text': 'panda'}, tool_call_id='...')]
```

**Run them yourself:**
```python
results = response.execute_tool_calls()
# [ToolResult(name='upper', output='PANDA', tool_call_id='...')]
```

**Continue the conversation in one call:**
```python
follow_up = response.reply()
print(follow_up.text())
```

Zero-arg `reply()` on a response that made tool calls auto-executes them and feeds the results back. Pass an explicit `tool_results=` to skip auto-execution and use your own (e.g. mutated, mocked, or pre-computed) results. Pass a `prompt=` string to also append a user turn after the tool results.

Async `reply()` is awaitable (`await response.reply()`) so the auto-execute path can `await` `execute_tool_calls()` internally.

---

## Persisting and continuing conversations

Two new round-trip methods:

```python
payload = response.to_dict()                      # JSON-safe dict
restored = llm.Response.from_dict(payload, model=model)
follow_up = restored.reply("Add 3 to that")       # continues from disk
```

The `to_dict()` payload captures everything needed: model id, the input chain (`prompt.messages`), the assembled output (`response.messages()`), and any options. The async equivalents are `AsyncResponse.from_dict()` / `await async_response.reply(...)`.

Provider-specific opaque data (Anthropic thinking signatures, Gemini thought signatures, OpenAI encrypted reasoning blobs) round-trips automatically via `Part.provider_metadata` so multi-turn continuity Just Works across save/load.

---

## At a glance

| Capability                        | Before                            | After                                                 |
|-----------------------------------|-----------------------------------|-------------------------------------------------------|
| Iterate response                  | `for chunk in response`           | unchanged (yields text strings)                       |
| Get full text                     | `response.text()`                 | unchanged                                             |
| Get assembled output              | —                                 | `response.messages()` (list of `Message`)             |
| Get streaming events              | —                                 | `response.stream_events()` / `astream_events()`       |
| Distinguish reasoning from text   | —                                 | `event.type == "reasoning"` vs `"text"`               |
| Mark opaque reasoning             | —                                 | `ReasoningPart(text="", redacted=True)`               |
| Run tools manually                | —                                 | `response.execute_tool_calls()`                       |
| Continue a conversation manually  | manually rebuild `messages=`      | `response.reply(...)` — auto-runs tools if present    |
| Save and restore                  | hand-rolled                       | `response.to_dict()` / `Response.from_dict()`         |
| Round-trip provider tokens        | each plugin's own ad-hoc storage  | `Part.provider_metadata` (namespaced by provider)     |

---

## Notes for upgrading

- **String yields still work** in plugins. Anything that returned plain `str` from `execute()` continues unchanged.
- **`for chunk in response`** still yields text strings only — non-text events are filtered out of the iterator.
- **`response.text()`** still concatenates the visible text answer.
- **`response.messages()`** is the new way to access the structured assistant turn(s). It auto-forces execution; on async responses, `await response.messages()`.
- **`model.chain(...)`** API is unchanged; the chain still drives the multi-turn tool loop. New: it correctly carries `system=` / `system_fragments=` through tool-result turns.
- **`response.reply(...)`** is new — it replaces the old "manually build messages= and call model.prompt" pattern.

---

## Plugin authoring (brief)

Plugin authors get a richer contract:

- Yield `StreamEvent` objects (or plain `str` for the simple case).
- Read `prompt.messages` (the canonical input chain) instead of walking conversation history.
- Stash opaque provider data in `StreamEvent.provider_metadata`; the framework merges it onto the finalized Part and round-trips it via `to_dict()`.
- For opaque reasoning, yield `StreamEvent(type="reasoning", chunk="", redacted=True)` once — no side-channel attribute needed.

See `docs/plugins/advanced-model-plugins.md` for the full plugin authoring guide.

simonw:
In an earlier branch ([dag](https://github.com/simonw/llm/commits/dag/)) I experimented with turning the SQLite logs into a graph structure and recording all of the parts - I think we can land and ship this in an alpha *without* any logging changes, other than I want to log reasoning text separately such that it can be shown in the Markdown log export from `llm logs`.

simonw:
I prototype documenting the serialization format but I don't think that is necessary yet, that can happen as part of the upcoming logging work.

simonw:
Here are two test scripts for the Python API and the LLM CLI tool against any model: https://gist.github.com/simonw/3a4935e8e290547ca2eb38a1e97c11fa

To run the Python one:

```
uv run https://gist.githubusercontent.com/simonw/3a4935e8e290547ca2eb38a1e97c11fa/raw/e5843dadd390ec2f9f13d84d6e9ac3cf9f112826/exercise_032_features.py \
  gpt-5.4-nano
```

And for the shell script:

```
wget https://gist.githubusercontent.com/simonw/3a4935e8e290547ca2eb38a1e97c11fa/raw/e5843dadd390ec2f9f13d84d6e9ac3cf9f112826/exercise_032_cli.sh
chmod 755 exercise_032_cli.sh
./exercise_032_cli.sh gpt-5.4-nano
```

llm/default_plugins/openai_models.py
```diff
@@ -970,8 +1009,30 @@ def execute(
...
+                )
+        # Read reasoning_tokens from usage BEFORE set_usage runs —
+        # set_usage pops top-level keys and passes the rest through
+        # simplify_usage_dict, which strips zero-valued entries.
+        if usage:
+            reasoning_tokens = (usage.get("completion_tokens_details") or {}).get(
+                "reasoning_tokens", 0
+            )
+            if reasoning_tokens:
+                yield StreamEvent(type="reasoning", chunk="", redacted=True)
```
simonw: Pretty sure we don't need this.

simonw:
I think this is good enough for an alpha.

PR #1434, opened 2026-05-05
Experimental prompt.display_reasoning mechanism

simonw:
So you can turn off display reasoning and compatible plugins can then opt not to request reasoning summaries from their models, even as they do actually run reasoning.

Not yet sure if the design is right.


simonw:
I'm doing this as `hide_reasoning=True` instead.

PR #1435, opened 2026-05-06
OpenAI Responses API model classes

simonw:
> We need to leave the existing chat completion class alone, since other plugins import it
> 
> Instead we will add a new class which implements the responses API but also offers a special chat_completions Boolean option which, if set, dispatches to the existing chat completions implantation. This is so when we register model they can use the new responses class but callers can still opt in to chat completions using "-o chat_completions 1" if they really need to
> 
> Start by building that new class and testing it interest against the real API - there is an OPENAI_API_KEY in this environment so testing with "llm -m gpt-5.5 hello" should work, try that once first to confirm it does before you start coding
> 
> Get it all working and then add new tests by capturing test data using the mechanism described in the README

https://claude.ai/code/session_01S7UevnY7zKogbthSz4UXJv


simonw:
I had Codex run a manual testing run against the branch using Showboat, results here: https://gist.github.com/simonw/19cfef23659a939c12b1f8a6ccfe227a

simonw:
Had Codex write and execute a script to test ALL the model IDs, which I used to help decide which ones to port to Responses: https://gist.github.com/simonw/4d1a1749475b092a328b92ae5f272e87

PR #1442, opened 2026-05-12
--hide-reasoning and hide_reasoning=True parameters

simonw:
Previously we had a `-R/--no-reasoning` option to `llm prompt` which was misleadingly named - it only suppressed output of reasoning text, it did not turn reasoning off - and was not passed through to the models.

As of #1435 the default OpenAI model plugin requests `"summary": "auto"` reasoning summaries. You can hide these with `-R` but they are still requested.

This change passes `hide_reasoning=True` through on the `Prompt` object, such that plugins can opt not to request visible reasoning tokens.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -32,7 +32,7 @@ Plugin authors should read the expanded {ref}`Advanced model plugins <advanced-m
 ### CLI
 
 - `llm prompt` and `llm chat` now display visible reasoning text to stderr in a dim style while the response streams.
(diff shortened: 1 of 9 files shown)


simonw:
- Docs for plugin implementors: https://github.com/simonw/llm/blob/e020d7c8c18425200addd5b91af76c176691051f/docs/plugins/advanced-model-plugins.md#reasoning-tokens
- End-user Python API docs: https://github.com/simonw/llm/blob/2acc0aa64974316e491d5008a6d1f360ddc2f98d/docs/python-api.md#hiding-reasoning-output

PR #1454, opened 2026-05-21
Keep monotonic_ulid monotonic when the system clock moves backward

c-tonneslan:
`monotonic_ulid()` documents that its result is "strictly larger than every other ULID returned by this function inside the same process", and that it "works the same way the reference JavaScript `monotonicFactory` does".

It doesn't, quite. It only handles `now_ms == last_ms` as the increment-the-randomness case:

```python
if now_ms == last_ms:
    # increment randomness, keep last timestamp
    ...
# New millisecond, start fresh
_last = _fresh(now_ms)
```

When the system clock moves **backward** — an NTP correction, a VM migration, a manual change — `now_ms < last_ms`, which skips that branch and falls through to `_fresh(now_ms)`. That builds a ULID whose timestamp prefix is *smaller* than the previous one, so it sorts before it and the monotonicity guarantee is broken.

The JS `monotonicFactory` this is modelled on uses `seed <= lastTime` precisely to cover this case. This change does the same: when the clock hasn't moved forward, keep the previous timestamp and increment the randomness.

Added `test_monotonic_ulid_survives_backward_clock`, which patches `time.time_ns` to jump backward between calls. It fails on `main` (the post-jump ULID sorts first) and passes with the fix. Existing `test_monotonic_ulids` and the rest of `test_utils.py` still pass.

### DIFF
--- a/llm/utils.py
+++ b/llm/utils.py
@@ -692,7 +692,8 @@ def monotonic_ulid() -> ULID:
 
     It works the same way the reference JavaScript `monotonicFactory` does:
     * If the current call happens in the same millisecond as the previous
-        one, the 80-bit randomness part is incremented by exactly one.
+        one, or the system clock has moved backward, the 80-bit randomness
+        part is incremented by exactly one.
     * As soon as the system clock moves forward, a brand-new ULID with
         cryptographically secure randomness is generated.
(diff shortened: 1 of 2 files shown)


simonw:
Closed in favor of https://github.com/simonw/llm/pull/1641

PR #1480, opened 2026-06-09
Tool implementations can receive the ToolCall via llm_tool_call param

simonw:
Tool functions (sync or async, including Toolbox methods) that declare a parameter named `llm_tool_call` are now passed the `llm.ToolCall` object for the current invocation. The parameter is reserved: it is excluded from the input schema exposed to the model and is only injected when declared explicitly - a `**kwargs` catch-all does not receive it.

This lets tool implementations key external state against the unique `tool_call_id`, e.g. for human-in-the-loop approval flows that need to resume a specific tool call after the answer arrives.

### DIFF
--- a/docs/python-api.md
+++ b/docs/python-api.md
@@ -211,6 +211,26 @@ response = model.chain(
 print(response.text())
 ```
 
+(python-api-tools-llm-tool-call)=
+
+#### Accessing the tool call from inside a tool
+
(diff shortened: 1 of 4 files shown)


simonw:
Claude Fable 5 suggested this one while working on adding human-in-the-loop features to Datasette Agent.

PR #1482, opened 2026-06-09
PauseChain primitive + chain resume from pending tool calls

simonw:
I need this for Datasette Agent, so it can pause a tool call in the middle in order to ask for human confirmation in a way that survives a server restart.

> Two features that together give chains a first-class suspend/resume story for human-in-the-loop tools:
> 
> llm.PauseChain: raise inside a tool implementation to stop the chain cleanly. Unlike other exceptions it is not converted into an error ToolResult - it propagates to the caller with .tool_call (the paused call) and .tool_results (completed sibling results) attached, and no provider call is made with a placeholder result. Failure semantics for concurrent tool execution are now defined: async sibling tasks always run to completion before a pause or hook exception propagates (gather with return_exceptions, raised after collection), so nothing is orphaned mid-flight; sync execution stops at the paused call, leaving later calls unstarted so they can safely run on resume.
> 
> Chain resume: chain(messages=history, tools=...) now detects a history ending in an assistant message with unresolved tool calls - e.g. one persisted when a previous run paused or crashed - executes those calls through the normal before_call/after_call machinery (skipping any that already have results), then sends the results to the model as a standard tool-result turn. A resumed call may pause again, enabling multi-question flows. Histories where a user or assistant message follows the calls are left alone. Also adds execute_tool_calls(tool_calls_list=) for executing an explicit list.

Follows:
- #1481 
- #1480


simonw:
Here's the code simplification in Datasette Agent this enabled:

https://github.com/datasette/datasette-agent/commit/01eed730d4ba9c1c352afd3b4bf0a8d4cc85d63e

simonw:
Huh, Fable wrote very useful documentation for this: https://github.com/simonw/llm/blob/55ace5d6a8318fc54614dc1e5db0216f7a714e90/docs/python-api.md#pausing-a-chain-from-inside-a-tool

PR #1485, opened 2026-06-10
Fix tool_instances attribution for multi-toolbox runs (#1398)

### LINKED ISSUE #1398: _BaseResponse.log_to_db() uses stale loop variable -- multi-tool runs log wrong toolbox metadata
### Bug

In `_BaseResponse.log_to_db()`, the `tool_instances` INSERT inside the `for tool_result in self.prompt.tool_results` loop references `tool.plugin` and `tool.name` from a previous `for tool in self.prompt.tools` loop. In Python, the loop variable retains its last value after the loop ends, so all tool results are attributed to whichever toolbox was last in `self.prompt.tools`.

This only affects Toolbox-backed tools (not plain function tools). There appears to be existing evidence in `tests/test_plugins.py` where `Memory_*` results are attributed to the `Filesystem` toolbox.

### Root cause

In `llm/models.py`:

```python
for tool in self.prompt.tools:
    # ... inserts tool rows ...

# After this loop, tool = last element of self.prompt.tools

for tool_result in self.prompt.tool_results:
    # Uses tool.plugin and tool.name here
    # But tool is stale -- always the last tool from the previous loop
    db["tool_instances"].insert({
        ...,
        "plugin": tool.plugin,
        "name": tool.name,
        ...
    })
```

### Impact

Database logs for multi-toolbox runs contain incorrect tool attribution. Debugging and auditing based on the `tool_instances` table will be misleading.

### Suggested fix

Look up the correct tool for each `tool_result` by matching on the tool call's name/ID rather than relying on the leaked loop variable.

(Found during a multi-LLM code review using [sqry](https://github.com/nicholasgasior/sqry) AST analysis + Codex + Gemini cross-validation.)

riteshkew:
Fixes #1398.

In `_BaseResponse.log_to_db()`, the `tool_instances` insert inside the `for tool_result in self.prompt.tool_results` loop was reading `tool.plugin` and `tool.name` from the previous `for tool in self.prompt.tools` loop variable, which retains the last tool after the loop ends. Every toolbox-backed tool result was attributed to whichever tool happened to be last in `self.prompt.tools`.

The fix builds `tools_by_name` next to the existing `tool_ids_by_name` and looks up the correct tool by `tool_result.name` when inserting the `tool_instances` row. If no matching tool is found we skip the insert (instead of writing a wrong row from the leaked variable).

`tests/test_plugins.py::test_toolbox_logging_async` was asserting the buggy attribution (Memory_set/Memory_get → `Filesystem`); updated to expect Memory_set/Memory_get → `Memory`. Full test suite still passes (782 passed).

### DIFF
--- a/llm/models.py
+++ b/llm/models.py
@@ -1529,9 +1529,11 @@ class _BaseResponse:
 
         # Persist any tools, tool calls and tool results
         tool_ids_by_name = {}
+        tools_by_name = {}
         for tool in self.prompt.tools:
             tool_id = ensure_tool(db, tool)
             tool_ids_by_name[tool.name] = tool_id
+            tools_by_name[tool.name] = tool
             db["tool_responses"].insert(
                 {
                     "tool_id": tool_id,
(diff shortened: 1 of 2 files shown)


riteshkew:
One design note worth flagging: when a `tool_result.name` has no matching tool in `self.prompt.tools` (e.g. a result from a tool that was removed before logging), the old code wrote a row with the stale loop-variable's plugin/name. This PR makes that path silently skip the `tool_instances` insert instead — the row simply doesn't exist. If you'd rather fall back to writing the row with `tool_result.name` as the instance name (or raise), happy to change it.

PR #1496, opened 2026-06-21
fix: follow redirects and strip charset in URL attachment handling

### LINKED ISSUE #1046: Ollama OCR with URLs fails where filepaths succeed
Really weird bug. I wrote about this here: https://simonwillison.net/2025/May/18/qwen25vl-in-ollama/

```bash
llm install llm-ollama
ollama pull qwen2.5vl

llm -a https://static.simonwillison.net/static/2025/poster.jpg \
  'convert to markdown' -m qwen2.5vl
```
I got back:

> It looks like the image you provided is a jumbled and distorted text, making it difficult to interpret. If you have a specific question or need help with a particular topic, please feel free to ask, and I'll do my best to assist you!

But... the same exact prompt after downloading the file gives a much better result:
```bash
wget https://static.simonwillison.net/static/2025/poster.jpg
llm -m qwen2.5vl 'convert to markdown' -a poster.jpg
```

xbrxr03:
## Summary

Fixes #1046

When attaching images via URL, two bugs caused failures:

### Bug 1: HTTP redirects not followed
`httpx.head()` and `httpx.get()` in `Attachment.resolve_type()` and `Attachment.content_bytes()` did not follow HTTP redirects (301/302). Any image URL that redirects — CDNs, short links, etc. — would fail instead of fetching the actual image.

This is inconsistent with the fragment loader in `cli.py` which already uses `follow_redirects=True` (line 161).

### Bug 2: Content-Type charset params not stripped
HTTP responses may return `Content-Type: image/jpeg; charset=utf-8`, but this full string was stored as the attachment type. When checked against model `attachment_types` sets like `{"image/jpeg", "image/png", ...}`, the charset-augmented string would not match, causing errors like:

```
This model does not support attachments of type image/jpeg; charset=utf-8
```

### Changes

**`llm/models.py`**
- `Attachment.resolve_type()`: Added `follow_redirects=True` to `httpx.head()` and stripped charset params from Content-Type header
- `Attachment.content_bytes()`: Added `follow_redirects=True` to `httpx.get()`

**`llm/cli.py`**
- `resolve_attachment()`: Added `follow_redirects=True` to `httpx.head()` and stripped charset params from Content-Type header

### Testing

8 new tests in `tests/test_attachment_url_fixes.py`:
- 3 tests for charset stripping in `Attachment.resolve_type()`
- 2 tests for redirect following in `Attachment` class methods
- 3 tests for charset stripping and redirect following in `resolve_attachment()`

All existing tests continue to pass.

### Note on the original issue

The original report (#1046) described different OCR results between URL and filepath attachments with Ollama. Subsequent investigation by @andrejartuchov-prog found the bytes were identical between both paths, suggesting qwen2.5-VL nondeterminism rather than an attachment bug. However, this PR fixes two real, reproducible bugs in URL attachment handling that would cause failures for redirected URLs and servers returning charset params.

### DIFF
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -263,9 +263,12 @@ def resolve_attachment(value):
     if "://" in value:
         # Confirm URL exists and try to guess type
         try:
-            response = httpx.head(value)
+            response = httpx.head(value, follow_redirects=True)
             response.raise_for_status()
-            mimetype = response.headers.get("content-type")
+            mimetype = response.headers.get("content-type", "")
+            # Strip parameters (e.g. "charset=utf-8") so "image/jpeg; charset=utf-8"
+            # becomes "image/jpeg" for consistent matching against attachment_types.
+            mimetype = mimetype.split(";")[0].strip()
         except httpx.HTTPError as ex:
             raise AttachmentError(str(ex))
         return Attachment(type=mimetype, path=None, url=value, content=None)
--- a/llm/models.py
+++ b/llm/models.py
@@ -91,9 +91,12 @@ class Attachment:
         if self.path:
             return mimetype_from_path(self.path)
         if self.url:
-            response = httpx.head(self.url)
+            response = httpx.head(self.url, follow_redirects=True)
             response.raise_for_status()
-            return response.headers.get("content-type")
+            content_type = response.headers.get("content-type", "")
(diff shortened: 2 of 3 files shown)


xbrxr03:
Closing — 17 days with zero maintainer engagement. Per our pipeline rule: no response after 7 days, we close.

Issue #1445, opened 2026-05-15
register_models should skip model registration when no key is configured


antisilent:
Currently, the built-in OpenAI plugin registers ~50 models unconditionally at startup, regardless of whether an OpenAI API key is present. This causes several problems:

- llm models output is cluttered with models the user cannot use
- Model IDs like gpt-4o-mini collide with identically-named models from other plugins (e.g. a LiteLLM proxy that exposes the same names)
- The default model resolves to an OpenAI model, so bare llm invocations fail with No key found for users who don't use OpenAI at all

The fix is minimal. llm.get_key already returns None when no key is available, so:

```python
def register_models(register):
    if not llm.get_key(alias="openai", env="OPENAI_API_KEY"):
        return
    # rest unchanged
```

This is also the right contract for any plugin — if credentials aren't configured, the plugin should be invisible. The same pattern would apply to an extracted llm-openai plugin (see #335).

PR #1498, opened 2026-06-24
Only register bundled OpenAI models when an OpenAI key is configured

gauravmm:
## What

The default OpenAI plugin always registered its bundled models (`gpt-4o`, `gpt-5.x`, `o`-series, ~50 of them) even when no OpenAI API key is configured, so `llm models` listed models that can never actually be used.

This gates the bundled registrations behind a key check — present means either the `openai` entry in `keys.json` or the `OPENAI_API_KEY` env var. With no key, the bundled OpenAI models are hidden from `llm models`.

Models added via `extra-openai-models.yaml` **always** load regardless, since they often target other providers / their own keys.

## How

In `register_models()`, when no OpenAI key resolves, the local `register` is swapped for a no-op for the bundled section only; the original is kept as `_register` and used for the `extra-openai-models.yaml` loop.

## Tests

- `tests/conftest.py`: the suite exercises the bundled models, so the autouse fixture now sets a dummy `OPENAI_API_KEY`. Tests for the no-key path delete it explicitly.
- New `test_bundled_openai_models_require_key` (env / keys.json / none) and `test_extra_openai_models_load_without_key`.
- Full suite passes; `black --check` and `ruff check` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/openai-models.md
+++ b/docs/openai-models.md
@@ -17,6 +17,8 @@ llm keys set openai
 ```
 Then paste in the API key.
 
+The bundled OpenAI models are only listed by `llm models` once an OpenAI key has been configured this way (or via the `OPENAI_API_KEY` environment variable). Without a key they are hidden, since they cannot be used. Models you add through {ref}`extra-openai-models.yaml <openai-extra-models>` are always listed regardless.
+
 (openai-models-language)=
 
 ## OpenAI language models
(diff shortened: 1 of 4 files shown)


gauravmm:
Turns out this duplicates work in #1445 and #1446, so I'm closing it in favor of the earlier ones.

PR #1507, opened 2026-07-02
Preserve streaming timings from OpenAI-compatible providers.

dmsnell:
llama.cpp's OpenAI-compatible streaming endpoint can include token/sec metrics in a top-level `timings` field on the final streamed chunk. `llm` was already preserving usage from that chunk, but `combine_chunks()` discarded provider-specific timing metadata, so `llm logs --json` could not show the throughput data even when the provider returned it.

In this patch, `timings` is captured while combining streamed chunks and include it in the persisted response JSON. Add a regression test using a llama.cpp-style streaming response.

Co-Authored-By: Codex (GPT-5.5)

### DIFF
--- a/llm/default_plugins/openai_models.py
+++ b/llm/default_plugins/openai_models.py
@@ -1986,8 +1986,12 @@ def combine_chunks(chunks: List) -> dict:
     # those later on
     logprobs = []
     usage = {}
+    provider_extras = {}
 
     for item in chunks:
+        item_dump = item.model_dump()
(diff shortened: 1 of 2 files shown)


dmsnell:
I’ve seen this before, where `llm` would show no timing information from OpenAI-compatible API calls, but it wasn’t until working with `llama.cpp` locally where I examined the streaming chunks that I discovered that the information was there, but `llm` was discarding it.

This returns the API timing information into the logs.

PR #1512, opened 2026-07-04
Fix fragment filter using OR instead of EXISTS...UNION

### LINKED ISSUE #1511: llm logs -f fragment filter returns no results on SQLite 3.51.0+
I was poking around a fresh clone of `llm` and ran the test suite, and ran into 5 failures in `tests/test_llm_logs.py`, all related to the `-f`/`--fragment` filter on `llm logs`. Dug into it and it turns out this is a genuine SQLite behavior issue, not a flaky test.

The fragment filter builds SQL like this (in `logs_list`, `cli.py`):

```sql
exists (
    select 1 from prompt_fragments
    where prompt_fragments.response_id = responses.id
    and prompt_fragments.fragment_id in (select fragments.id from fragments where hash = :f0)
    union
    select 1 from system_fragments
    where system_fragments.response_id = responses.id
    and system_fragments.fragment_id in (select fragments.id from fragments where hash = :f0)
)
```

On SQLite 3.51.0 (what I've got bundled with Python on this machine), this returns zero rows whenever *both* `prompt_fragments` and `system_fragments` have at least one row matching the hash somewhere in the table - even if only one of them actually matches the specific outer `responses.id` being checked. In other words, `EXISTS` wrapping a `UNION` of two correlated subqueries seems to break the correlation once both branches are "live". I couldn't find this documented anywhere, so it might just be a quirk of a very recent SQLite version.

Here's a minimal repro with nothing but the stdlib, no llm/sqlite-utils involved:

```python
import sqlite3
con = sqlite3.connect(":memory:")
con.execute("create table R (id integer primary key)")
con.execute("create table A (response_id integer, val text)")
con.execute("create table B (response_id integer, val text)")
con.executemany("insert into R values (?)", [(1,), (2,), (3,)])
con.execute("insert into A values (1, 'x')")   # R1 should match via A
con.execute("insert into B values (2, 'y')")   # R2 matches via B, different val
con.commit()

sql = """
select R.id from R
where exists (
    select 1 from A where A.response_id = R.id and A.val = :v
    union
    select 1 from B where B.response_id = R.id and B.val = :v
)
"""
print(con.execute(sql, {"v": "x"}).fetchall())
# expected [(1,)], actually prints []
print(sqlite3.sqlite_version)  # 3.51.0 here
```

Rewriting the two `EXISTS` branches as `(exists (...) or exists (...))` instead of one `EXISTS` wrapping a `UNION` fixes it for me and is semantically the same thing. I've got that change ready plus a full test run (760 passed) - will open a PR against this issue shortly.

Cyrus580529:
Fixes #1511.

The `-f`/`--fragment` filter on `llm logs` was building a single `EXISTS(...)` wrapping a `UNION` of two correlated subqueries (one against `prompt_fragments`, one against `system_fragments`). On SQLite 3.51.0 that combination silently drops matches once both tables have a matching row somewhere, even if only one of them actually correlates to the response being checked - see the issue for a minimal repro that has nothing to do with this codebase.

The fix here just swaps it for `(exists (...) or exists (...))`, which is the same logic without going through a compound `SELECT` inside the `EXISTS`. No behavior change intended beyond "it actually filters correctly now."

Before this change, `tests/test_llm_logs.py` had 5 failing tests on my machine (all fragment-filter related). After the change the full suite passes: 760 passed, 1 skipped, 8 xfailed, 13 xpassed.

### DIFF
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -1780,19 +1780,22 @@ def logs_list(
 
         for i, fragment_hash in enumerate(fragment_hashes):
             exists_clause = f"""
-            exists (
-                select 1 from prompt_fragments
-                where prompt_fragments.response_id = responses.id
-                and prompt_fragments.fragment_id in (
-                    select fragments.id from fragments
-                    where hash = :f{i}
+            (
+                exists (
+                    select 1 from prompt_fragments
+                    where prompt_fragments.response_id = responses.id
+                    and prompt_fragments.fragment_id in (
+                        select fragments.id from fragments
+                        where hash = :f{i}
+                    )
                 )
-                union
-                select 1 from system_fragments
-                where system_fragments.response_id = responses.id
-                and system_fragments.fragment_id in (
-                    select fragments.id from fragments
-                    where hash = :f{i}
+                or exists (
+                    select 1 from system_fragments
+                    where system_fragments.response_id = responses.id
+                    and system_fragments.fragment_id in (
+                        select fragments.id from fragments
+                        where hash = :f{i}
+                    )
                 )
             )
             """


simonw:
I'm going to use the fix from:
- #1571

PR #1525, opened 2026-07-12
fix: do not commit inside the content_hash embeddings migration

### LINKED ISSUE #1523: Embeddings migrations fail with sqlite-utils 4.0; embed-multi masks it as "You need to specify an embedding model"
With sqlite-utils 4.0 installed, the embeddings migrations fail on every released version of llm — and `llm embed-multi` reports the failure as **"You need to specify an embedding model (no default model is set)"** even when `-m` is explicitly given, which sent me debugging model configuration when the actual problem was a migration crash.

Three parts, each reproducible with current releases (Python 3.12, sqlite-utils 4.0):

## 1. llm 0.31.1 (latest stable): `m003_add_updated` uses `db.query()` for an UPDATE

```python
import sqlite_utils
from llm.embeddings_migrations import embeddings_migrations

db = sqlite_utils.Database("fresh.db")
embeddings_migrations.apply(db)
```

```
File ".../llm/embeddings_migrations.py", line 35, in m003_add_updated
  db.query(
File ".../sqlite_utils/db.py", line 868, in query
  raise ValueError(message)
ValueError: query() can only be used with SQL that returns rows - use execute() for other statements
```

Commit 67adad2c ("Use db.execute() not db.query()", refs simonw/sqlite-utils#758) already fixes this on `main`, but it isn't in any release — 0.31.1 was tagged after that commit landed and still ships `db.query()`, and 0.32a3 predates it.

## 2. llm main (67adad2c): `m004_store_content_hash` still fails — `with db.conn:` inside the migration runner's transaction

Same script with `llm @ git+https://github.com/simonw/llm`:

```
File ".../sqlite_utils/migrations.py", line 132, in apply
  with db.atomic():
File ".../sqlite_utils/db.py", line 631, in atomic
  self.conn.execute("COMMIT")
sqlite3.OperationalError: cannot commit - no transaction is active
```

sqlite-utils 4's `Migrations.apply()` wraps each migration in `db.atomic()`. `m004_store_content_hash` does `with db.conn:` around its UPDATEs; the stdlib sqlite3 connection context manager commits the enclosing transaction on exit, so `atomic()`'s own `COMMIT` then has no transaction to commit. Dropping the `with db.conn:` (the runner already provides the transaction) or using a savepoint would fix it.

## 3. `embed-multi` masks any migration failure as a missing-model error

```
$ printf 'id,name\n1,hello\n2,world\n' > data.csv
$ llm embed-multi demo data.csv --format csv -d demo.db -m ada-002
Error: You need to specify an embedding model (no default model is set)
```

A model **is** specified. The cause is the broad `except ValueError` around the `Collection(...)` constructor in `cli.py` (line ~3438 on main) — `Collection.__init__` applies the migrations, so the migration `ValueError` from part 1 is caught and re-raised as the missing-model message. Narrowing that handler (or letting migration errors propagate) would have surfaced the real traceback immediately.

Happy to provide more detail — I hit this embedding a ~2,000-image collection with llm-clip and initially spent the session debugging model registration because of the message in part 3.

Ewertonslv:
Closes #1523.

## The bug

Every embeddings command fails on a fresh install today:

```
sqlite3.OperationalError: cannot commit - no transaction is active
```

`pyproject.toml` allows `sqlite-utils>=3.37` and `sqlite-migrate>=0.1a2`, so a new install now resolves to sqlite-utils 4.1 and sqlite-migrate 0.2 — and 0.2 is just a shim that re-exports `sqlite_utils.Migrations`.

That matters because `sqlite_utils.Migrations.apply()` runs each migration inside `db.atomic()`, i.e. `BEGIN` ... `COMMIT`. `m004_store_content_hash` wrapped its two backfill `UPDATE`s in `with db.conn:`, and the stdlib connection context manager commits on exit — so it ended the transaction `atomic()` had opened, and `atomic()`'s own `COMMIT` then had nothing left to commit.

This is why `test_embed.py` is currently red on `main`.

## The fix

Drop the `with db.conn:` wrapper and let the enclosing transaction do its job — the same shape as 67adad2, which fixed `m003_add_updated` for sqlite-utils 4.

I deliberately did not use `db.atomic()` here: it doesn't exist in sqlite-utils 3.x, which `pyproject.toml` still supports.

## Verification

- New test `test_embeddings_migrations_do_not_commit` fails on `main` with the exact `OperationalError` above and passes with the fix. It's skipped on sqlite-utils 3.x, where migrations already commit internally (`transform()` does), so the invariant cannot hold there and the bug cannot occur.
- Full suite green on **both** stacks: 765 passed on sqlite-utils 4.1 (currently red on `main`), 764 passed + 1 skipped on 3.39.
- Checked the backfill is still durable, not just crash-free: applied the migrations to a file-backed DB with rows seeded before `m004`, reopened it on a fresh connection, and confirmed `content_hash` is persisted — on 4.1 and on 3.39.
- `black --check .`, `mypy llm` and `ruff check .` all pass.

## One thing I left alone

#1523 also reports `embed-multi` masking the failure as *"You need to specify an embedding model"*. That's the broad `except ValueError` around `Collection(...)` in `cli.py` — but it does not catch this crash, which is an `OperationalError`. The `ValueError` the reporter hit came from `m003`'s old `db.query()` call, already fixed in 67adad2. Narrowing that `except` is a real question but a separate judgement call, so I've left it for you.

### DIFF
--- a/llm/embeddings_migrations.py
+++ b/llm/embeddings_migrations.py
@@ -62,17 +62,16 @@ def m004_store_content_hash(db):
     db.conn.create_function("temp_md5", 1, md5)
     db.conn.create_function("temp_random_md5", 0, random_md5)
 
-    with db.conn:
-        db.execute("""
-            update embeddings
-            set content_hash = temp_md5(content)
-            where content is not null
-        """)
-        db.execute("""
-            update embeddings
-            set content_hash = temp_random_md5()
-            where content is null
-        """)
+    db.execute("""
+        update embeddings
+        set content_hash = temp_md5(content)
+        where content is not null
+    """)
+    db.execute("""
+        update embeddings
+        set content_hash = temp_random_md5()
+        where content is null
+    """)
 
     db["embeddings"].create_index(["content_hash"])
 
--- a/tests/test_migrate.py
+++ b/tests/test_migrate.py
@@ -117,6 +117,18 @@ def test_migrations_for_embeddings():
     assert db["embeddings"].foreign_keys[0].other_table == "collections"
 
 
+@pytest.mark.skipif(
+    not hasattr(sqlite_utils.Database, "atomic"),
+    reason="sqlite-utils 4 and higher run each migration inside a transaction",
+)
+def test_embeddings_migrations_do_not_commit():
+    # https://github.com/simonw/llm/issues/1523
+    db = sqlite_utils.Database(memory=True)
+    for migration in embeddings_migrations.pending(db):
+        with db.atomic():
+            migration.fn(db)
+
+
 def test_backfill_content_hash():
     db = sqlite_utils.Database(memory=True)
     # Run migrations up to but not including m004_store_content_hash


Ewertonslv:
`main` is green again, but I do not think the underlying issue is fixed. It looks like it is pinned around, and the new `>=4` job cannot reach it.

`1dc49af` changed `sqlite-migrate>=0.1a2` to `sqlite-migrate==0.1a2` in `pyproject.toml`. That pin is what makes the suite pass. The `test (ubuntu-latest, 3.14, >=4)` job added in #1558 installs `sqlite-utils>=4`, but `sqlite-migrate` stays at `0.1a2`, so the combination that actually breaks is never exercised.

I checked this on `5a4644a`, same tree and same virtualenv, changing one variable.

sqlite-utils 4.1.1 + sqlite-migrate 0.1a2, which is exactly what CI resolves to:

```
15 passed
```

sqlite-utils 4.1.1 + sqlite-migrate 0.2:

```
4 failed, 4 passed, 7 errors
sqlite3.OperationalError: cannot commit - no transaction is active
```

The mechanism: sqlite-migrate 0.2 is a shim that re-exports `sqlite_utils.Migrations`, and `Migrations.apply` in sqlite-utils runs each transactional migration inside `with db.atomic()`. `m004_store_content_hash` then opens `with db.conn:` at `llm/embeddings_migrations.py:66`, which commits when it exits and leaves the outer transaction with nothing to commit. sqlite-migrate 0.1x ships its own `Migrations` class with no `atomic()` wrapper, which is why the pin hides it.

#1523 is still open, and it reproduces for anyone who ends up with sqlite-migrate 0.2 installed. This PR removes the inner transaction so the pin could go back to `>=0.1a2`. If you would rather keep the inner block, marking that migration non-transactional is the other option. Either way I am happy to add `sqlite-migrate` to the CI matrix so the combination is actually covered.

Ewertonslv:
Closing this. The bug is already fixed on `main`, by a different change than the one here.

`774e62b` ("Drop sqlite-utils 3.x support, require 4.0 or higher") changed `m004_store_content_hash` to use `with db.atomic():` instead of `with db.conn:`. That solves the same problem this PR was solving. `148c19d` then removed the `sqlite-migrate` pin I wrote about in my previous comment, so the version combination that used to break is gone as well.

I checked this on `6afc1c4` before closing. On a clean `main` the migration tests pass:

```
$ pytest tests/test_migrate.py -q
6 passed
```

And when I put `with db.conn:` back, the existing suite catches it on its own, without the test from this PR:

```
$ pytest tests/test_migrate.py -q
FAILED tests/test_migrate.py::test_migrations_for_embeddings - sqlite3.OperationalError: cannot commit - no transaction is active
FAILED tests/test_migrate.py::test_backfill_content_hash - sqlite3.OperationalError: cannot commit - no transaction is active
2 failed, 4 passed
```

So the regression is already covered and this PR would not add anything. Sorry for the noise.

One note on #1523, which is still open. Parts 1 and 2 are fixed, but part 3 is not. In `llm/cli.py` (lines 3415-3422 on `6afc1c4`) the `embed-multi` command still wraps the `Collection(...)` call in a bare `except ValueError`, and reports any `ValueError` from it as "You need to specify an embedding model (no default model is set)". `Collection.__init__` runs `embeddings_migrations.apply(self.db)` before it checks the model, so a migration failure is still reported as a missing model even when `-m` was given. The `embed` command right above it already does this the safer way, by checking the model first and letting other errors through. I am happy to send a narrow fix if you want one.

PR #1540, opened 2026-07-19
docs: clarify model listing API key requirements

### LINKED ISSUE #1175: Enhance help tooltip for listing models
It would be great for the "help" tooltip on `llm models list` to inform the user that they need to set an API key before the models for a provider will be listed. I was confused as to why I couldn't see deepseek or openrouter models after installing those plugins, but I *could* see the Gemini ones.

lntutor:
## Summary

Clarifies the `llm models list --help` output so users know that some plugins only register models after an API key is configured. The help text points users to `llm keys set`, and a regression test covers both messages.

Fixes #1175.

## Testing

- `uv run pytest tests/test_llm.py -k "models_help_mentions_api_keys or llm_models_options"` (2 passed)
- `uv run ruff check llm/cli.py tests/test_llm.py`
- `uv run black --check llm/cli.py tests/test_llm.py`
- `git diff --check`

The full suite reached 701 passing tests but has existing embedding migration failures with `sqlite-utils 4.1.1` (`sqlite3.OperationalError: cannot commit - no transaction is active`), consistent with the open compatibility report in #1523. The changed model-list tests pass independently.

### DIFF
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -2497,7 +2497,11 @@ def render_model_with_options(model_id, *, async_=False):
 )
 @click.option("model_ids", "-m", "--model", help="Specific model IDs", multiple=True)
 def models_list(options, async_, schemas, tools, query, model_ids):
-    "List available models"
+    """List available models.
+
+    Some plugins only register their models after an API key is configured.
+    Run 'llm keys set' to store API keys for installed model providers.
+    """
     models_that_have_shown_options = set()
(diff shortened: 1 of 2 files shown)


lntutor:
Current head `843bfd70d4b3849daf4a98c929eef160cb9838f5` is mergeable, the Read the Docs preview is green, and there are no unresolved review threads on the current PR state. This looks ready for maintainer review when convenient.

PR #1542, opened 2026-07-19
Add count mode to logs command

### LINKED ISSUE #1071: llm logs --count should return a count of matching logs
This is technically a breaking change because `-n` has the longer form `--count` and is used to set the number of results to return.

I'm pre-1.0 so I'm going to break it!

It's really useful to be able to get a quick count for how many results are matching. For example:

```bash
llm logs --tools --count
```

lntutor:
## Summary

- keep `-n` for limiting displayed logs and rename its long form to `--limit`
- add `llm logs --count` using the existing query construction so all filters are respected
- document the new mode and update generated CLI help
- cover unfiltered, model, search, limit-independent, and tool-filtered counts

Closes #1071

## Validation

- `uv run pytest tests/test_llm_logs.py -q` (83 passed)
- `uv run black --check .`
- `uv run ruff check .`
- targeted Cog check for `docs/help.md` and `docs/logging.md`

The full suite reached 707 passing tests; its embedding migration failures reproduce unchanged on clean `upstream/main` with the locally resolved `sqlite-utils` version (`cannot commit - no transaction is active`).

### DIFF
--- a/docs/help.md
+++ b/docs/help.md
@@ -344,8 +344,9 @@ Usage: llm logs list [OPTIONS]
   Show logged prompts and their responses
 
 Options:
-  -n, --count INTEGER         Number of entries to show - defaults to 3, use 0
+  -n, --limit INTEGER         Number of entries to show - defaults to 3, use 0
                               for all
+  --count                     Output the number of matching log entries
   -d, --database FILE         Path to log database
   -m, --model TEXT            Filter by model or model alias
   -q, --query TEXT            Search for logs matching this string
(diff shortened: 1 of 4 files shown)


lntutor:
Rebased onto current `main` and resolved the log-query conflict by adapting `--count` to the new unified `merged_log_rows()` path, so it counts both current and legacy log storage while `--limit` continues to affect display only. Validation: all 87 tests in `tests/test_llm_logs.py` pass, and Ruff reports no issues for the changed Python files.

lntutor:
Current head `e3affe93742762b4c962a9bef87f87d314363e0d` is mergeable, the Read the Docs preview is green, and there are no unresolved review threads on the current PR state. The July 30, 2026 rebase/verification note on this PR still matches the current head. This looks ready for maintainer review when convenient.

PR #1561, opened 2026-07-26
response.execute_tool_call() method

### LINKED ISSUE #1560: Mechanism for custom execution of tool calls that triggers before/after callbacks
I was trying to build a plugin for the Apple foundation models using the https://github.com/apple/python-apple-fm-sdk library but ran into a problem: that library has its own tool execution mechanism which expects to be handed Python callables itself.

This broke our before/after tool mechanisms.

Solution: a `response.execute_tool_call()` method which those libraries can call - it dispatches the tool call through our stack.

simonw:
Refs:
- #1560

### DIFF
--- a/docs/plugins/advanced-model-plugins.md
+++ b/docs/plugins/advanced-model-plugins.md
@@ -380,6 +380,46 @@ response.add_tool_call(
 )
 ```
 
+### Provider-managed local tool calls
+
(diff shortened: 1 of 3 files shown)


simonw:
Documentation: https://github.com/simonw/llm/blob/fe830ef81869c097989fdd2a8d92753c26a16cae/docs/plugins/advanced-model-plugins.md#provider-managed-local-tool-calls

Issue #1314, opened 2025-12-05
Parts mechanism: conversations/responses needs to evolve to handle more complex conversation transcripts


simonw:
The current conversation/response/storage abstraction has some problems.

Gemini, Anthropic and OpenAI have been getting more sophisticated with how they handle longer conversations and tool calling. In  particular some of the more recent models can run tools server-side - like their web search tool - so the model does one or more tool calls before returning from the API.

These tool calls can even take place during the "reasoning" phase. The reasoning stuff is a problem because that can return text that is reasoning text that is not technically part of the final response but still deserves to be stored in SQLite and optionally made available to the Python API users.

Sometimes models return encrypted  blocks of additional context - these need to be stored too so they can be round-tripped to the API in future calls. This all adds up to much more complex requirements for how the conversation mechanism should work and how ongoing conversations should be persisted.

I brainstormed with Claude Opus 4.5 a bit, it was very useful: https://gistpreview.github.io/?8e871d0030a7b28fa40616d32611660e

simonw:
As pointed out in #1160 we have a problem right now where `llm logs -n 5` returns 5 most recent responses without any indication as to whether they were part of a conversation or not. That needs to be addressed by this new work as well.

simonw:
I think this work is also where we'll end up solving for annotations:
- #716

simonw:
Next job on this task will be to collect as many real-world JSON examples of complex transcripts that need to be logged from different API providers.

simonw:
Issues to consider with this:
- reasoning phases
- tool calls within reasoning v.s. tool calls later
- returned images
- returned audio
- maybe returned video in the future too?
- tool calls that happen server-side, like web search
- MCP can happen server-side too now

simonw:
Demo derived from https://cookbook.openai.com/examples/mcp/mcp_tool_guide

```bash
curl https://api.openai.com/v1/responses \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer $(llm keys get openai)" \
  -d '{
  "model": "gpt-4.1",
  "tools": [
    {
      "type": "mcp",
      "server_label": "gitmcp",
      "server_url": "https://gitmcp.io/openai/tiktoken",
      "allowed_tools": [
        "search_tiktoken_documentation",
        "fetch_tiktoken_documentation"
      ],
      "require_approval": "never"
    }
  ],
  "input": "search documentation for tokens"
}' | jq
```
Output:
```json
{
[...]

Issue #1092, opened 2025-05-26
llm -c does not carry over tools that use classes


simonw:
Spotted working on https://github.com/simonw/llm-tools-datasette
```bash
llm -T 'Datasette("https://datasette.io/content")' 'show tables' --td
llm -c 'Do something neat with the views'
```
This relates to:
- #1080

Need to log that initial setting in order to be able to reuse it on subsequent `llm -c` calls.

simonw:
I'm dropping this out of the stable release, see rationale here:
- https://github.com/simonw/llm/issues/1089#issuecomment-2910364659

simonw:
Update this when I implement this feature:

https://github.com/simonw/llm/blob/cc017a7f843ffc0c3481dcd97af4831d2df54759/docs/usage.md?plain=1#L211

PR #1562, opened 2026-07-26
Log new message shape to SQLite using content-addressed de-duplication

simonw:
The [0.32a](https://llm.datasette.io/en/latest/changelog.html#a0-2026-04-28) alpha series included a major rethink of the underlying data structures used within LLM.

- #1314

What was missing from those alphas was a way to store those in a round-trippable way in the SQLite `logs.db` database.

In addition to storing the new shapes, I want to be able to represent forked conversations without duplicating their entire history up to the point of the fork.

I also want to be able to implement an OpenAI chat completions style API endpoint - one where each round from the client carries the growing state of the existing conversation - without storing duplicates.

These last two requirements lead to content-addressable hashing, where chunks of content are stored once and linked into a tree based on the hash of that content.

I don't want to risk deleting or corrupting data from existing databases, so I'm going to implement this all as new tables. A new `turns` table represents the top-level turn (equivalent to the current `responses` table), then everything else is stored in terms of `messages` and `parts`.

Still to implement before landing this branch:

- [x] Ensure `model.prompt(prompt=...)` is correctly logged - there may be a bug where `messages=[...]` is correctly logged but the prompt from a `prompt=`) call is not
- [x] `llm logs` currently reads only from the old `responses` table, which is dual-written to. Decide on an approach and implement it.
- [x] Documentation for the new Python API for writing to a database, since this is meant to be a stable public API now.
- [x] I'm tempted to prototype an OpenAI chat completions server against this to prove the design is good. UPDATE: Did that in a new plugin: https://github.com/simonw/llm-chat-completions-server
- [x] What do we do about old data? We could backfill it, we could convert it on demand the first time it is accessed, we could leave it as-is and teach `llm logs` how to mix both new and old.
- [x] There is a `turn_fragments` table that is unused, `llm logs -f` should use it.
- [ ] `verify()` is N+1, should it exist? Should `llm logs verify` execute it?
- [ ] Improve `test_async_parity` tests.
- [x] Decide on if the experimental recursive CTE view in https://github.com/simonw/llm/blob/237f047859ac53f9b7335a7d10a2bf1b1feb8f97/conversation-views.sql becomes a feature.


simonw:
Deleting conversations is an interesting twist on this, because now that we have content-addressed deduplication we can't just delete rows without first checking they aren't part of some other conversation. We may need some kind of garbage collection mechanism here.

simonw:
The difference between `threads` and `turns` is non-obvious.

A **thread** is what we might call a conversation - it's effectively a pointer that indicates the tree that represents the most recent state of that conversation.

A **turn** captures an interaction with a model that may have created or updated a thread. Crucially this is where timestamps and which-model-records and token accounting live.

simonw:
I had Claude Code manually test this branch against `llm-gemini` and `llm-anthropic` which caught and fixed this async bug: 8e005ac8a79b202b7aac0c235f8394907fd7c692

simonw:
I think I **do** want the `llm logs` command to be able to both search and return logs from the older format, if we can make the code neat enough that it's not a burden to maintain in the future.

Maybe we could invent an intermediary format which the log rendering (and JSON log displaying) code handles, with separate code paths to load that from the old and new tables?

simonw:
For search, I want to be able to FTS against the original prompt (excluding fragments) and the text response - not the reasoning trace. That's how search works for the current system, I want the same thing against either threads or turns, need to decide which.

simonw:
Opus 5 was getting stressed out at running close to its context length without committing, so I let it commit a WIP with some failing tests and told it:

> `any more notes you need to leave for your successor? Stick them in notes-for-successor.md`

https://gist.github.com/simonw/1108699ee5e9211f53f7d6aec9e8138f

simonw:
I don't think `LogStore` should be a public API. The public API for this should be `Response.log_to_db()`, plus the schema itself.

Documenting the schema is one of the most important remaining tasks.

simonw:
CleanShot 2026-07-26 at 20 50 59@2x

There's one part of this design that I'm second-guessing right now: the JSON format used in the `parts` table.

Three illustrative examples of that JSON:

A regular piece of text:
```json
{"text":"This is version 0.32a3.","type":"text"}
```
A piece of text that embeds a reference to a fragment:
```json
{"text_ref":[{"fragment":1},{"literal":"\nWhat phrase repeats? Five words max."}],"type":"text"}
```
A tool call result:
```json
{"name":"llm_version","output":"0.32a3","tool_call_id":"bk95gnzr","type":"tool_result"}
```
I'm mainly not crazy about storing `{"text":"This is version 0.32a3.","type":"text"}` for every response text.

simonw:
One big gap in our work so far: we're no longer recording the configuration for a toolbox, so this:

```bash
llm -T 'SQLite("mydb.db") 'list tables'
```

Then
```bash
llm -c 'count rows in first table'
```

No longer works (using [llm-tools-sqlite](https://github.com/simonw/llm-tools-sqlite)).

Though actually it didn't work before!

```bash
uvx --with llm-tools-sqlite llm -T 'SQLite("data.db")' 'list tables'
uvx --with llm-tools-sqlite llm -c 'count rows in first table'
```
Outputs:
> `Error: Tool(s) SQLite_query, SQLite_schema not found. Available tools: llm_version, llm_time, SQLite`

simonw:
So all that's changed here is that we used to record (but not use) the configuration information, now we don't even store it.

simonw:
I'm adding a `text` column separate from `payload` which can be used for the common case of a response that is just text.

If there are fragments then `text` will be `null` and we will store the fragment mixed with literals in payload.

Also we will ditch the `"type"` key since that's already duplicated from a column.

simonw:
Also going to solve the `SQLite("path/to/db.db")` problem with this table:
```sql
create table tool_instantiations (
    tool_call_id text primary key,
    instance_name text,      -- "SQLite"
    instance_plugin text,    -- "llm_tools_sqlite"
    instance_arguments text  -- '{"path": "mydb.db"}'
);
```

simonw:
Here's what the new text column looks like:

CleanShot 2026-07-26 at 21 36 33@2x

docs/changelog.md
```diff
@@ -1,5 +1,16 @@
 # Changelog
 
+## Unreleased
+
+- `llm logs --json` now shows which configured toolbox instance served each tool result - the toolbox name, plugin and constructor arguments, so a `SQLite_query` result records that it ran against `SQLite("mydb.db")`. Recorded in the new `tool_instantiations` table for new logs, and read from the legacy `tool_instances` data for history logged by older versions of LLM, which recorded it without ever displaying it. [#1562](https://github.com/simonw/llm/pull/1562)
```
simonw: The bullets in this section should be ordered by most interesting first.

docs/logging.md
```diff
@@ -173,17 +173,26 @@ llm logs --cid 01h82n0q9crqtnzmf13gkyxawg
...
+```bash
+llm logs -q '"pet pelican"'
+```
+
 To switch to sorting with most recent first, add `-l/--latest`. This can be combined with `-n` to limit the number of results shown:
 ```bash
 llm logs -q 'cheesecake' -l -n 3
 ```
 
+Search spans both generations of log tables: new conversations are matched through the `turn_search` index over the content-addressed tables, and history recorded by older versions of LLM is matched through its original `responses_fts` index, with the two result sets merged. The relevance scores come from two separate indexes, so the interleaving between very old and new results is approximate.
```
simonw: No need to talk about details of relevance here, or details of tables. Just mention that search covers both old and new conversations.

docs/logging.md
```diff
@@ -276,6 +285,366 @@ llm logs backup /tmp/backup.db
...
+  part 0: type=text text='Now one for a pet walrus' payload=null
+
+assistant b2:a30a236e0d1b717c592e826d06e3c9d2
+  parent: b2:c785dd6c77540150c2647f406cacc76f
+  part 0: type=text text='Wallace. It pairs nicely with Percy.' payload=null
+
+turns:
+
+turn $TURN_1_ID
+  thread_id: $THREAD_ID
```
simonw: Instead of displaying `$TURN_1_ID` lets pick a baked in ULID and show that, it's easier to understand.

docs/logging.md
```diff
@@ -276,6 +285,366 @@ llm logs backup /tmp/backup.db
...
+- Each turn brackets one model call. Its `parent_message_hash` is the tip of the chain that was sent to the model - the last input message, usually that turn's user prompt - and its `tip_message_hash` is the chain tip after the model's reply was appended. The prompt and response that `llm logs` displays are derived by splitting the chain at the parent, rather than being stored a second time.
+- The thread's id is the conversation id, its name is derived from the first prompt and its `tip_message_hash` follows the head of the conversation as new turns are logged.
+- The ULID identifiers - sortable random identifiers issued in time order, the same id space older versions used for response ids - and the timestamp columns have been replaced with placeholders because they change on every run. The hashes have not: they depend only on the message content and its position in the chain, so replaying this conversation produces these exact four hashes.
```
simonw: This bit does not make sense. The reader doesn't need to know that we used to use ULIDs here and now we do not. They should know that we still use ULIDs for turns.

docs/logging.md
```diff
@@ -276,6 +285,366 @@ llm logs backup /tmp/backup.db
...
+
+### Content addressing as a contract
+
+A message's hash is calculated like this:
+
+1. Build the object `{"parent": parent_hash, "message": message}`, where `parent_hash` is the hash of the previous message in the chain (or `null` for the first message) and `message` is the message's dictionary representation - its role, its parts and any provider metadata.
+2. Serialize that object to canonical JSON: keys sorted, compact `,` and `:` separators with no extra whitespace, non-ASCII characters left unescaped.
+3. Hash the UTF-8 encoding of that string with [BLAKE2b](https://en.wikipedia.org/wiki/BLAKE_(hash_function)) using a 16 byte digest, and prefix the hex digest with `b2:`.
+
+The `b2:` prefix tags the hash with the algorithm that produced it. If a future version of LLM ever changes the digest or the canonical form, the change will be detectable instead of silently splitting the stored data into two incompatible halves.
```
simonw: This is over-explained.

docs/logging.md
```diff
@@ -276,6 +285,366 @@ llm logs backup /tmp/backup.db
...
+Attachments work the same way: the binary content lives in the `attachments` table, keyed by a SHA-256 hash of the bytes, and the part payload stores that id in place of the data.
+
+(logging-message-store-tables)=
+
+### Table by table
+
+The full schema for these tables appears in {ref}`the SQL schema section <logging-sql-schema>` below.
+
+- `messages` - one row per unique message. `hash` is the content address described above, `parent_hash` links to the previous message in the chain and `role` is `system`, `user` or `assistant`. `provider_metadata` holds any provider-specific data carried by the message; it participates in the hash.
+- `parts` - the content of each message, ordered by `position`. `text` holds the part's literal text when it borrows no fragments - raw and never parsed, so `select text from parts` reads as prose. `payload` holds whatever structure remains as JSON (fragment references, tool call fields, provider metadata), in the order the model produced it, or NULL when the text column carries the whole part. `type` and `tool_name` are their own columns for direct filtering; the type never appears inside the payload.
```
simonw: "when it borrows no fragments" is unclear.

docs/logging.md
```diff
@@ -276,6 +285,366 @@ llm logs backup /tmp/backup.db
...
+- `turn_fragments` - which fragments a turn was given, with their `kind` (`prompt` or `system`) and order. Provenance lives here rather than on the shared message rows, and this table is what powers `llm logs -f`.
+- `turn_search` - the searchable text of each turn: the literal prompt the user typed (fragment content excluded, via the `text_ref` literals described above) and the assistant's text output. An FTS5 index over this table, `turn_search_fts`, is what powers {ref}`llm logs -q <logging-search>`. Derived from the stored parts when a turn is logged; a turn with no prompt or response text, such as a pure tool call, gets no row.
+- `tool_instantiations` - which configured {ref}`toolbox <python-api-toolbox>` instance served a tool call: the toolbox name, its plugin and its constructor arguments, keyed by `tool_call_id`. Message rows are shared between conversations and so cannot carry this kind of local execution provenance; this table joins to the chain from outside it, and is what lets `llm logs --json` show that a `SQLite_query` call ran against `SQLite("mydb.db")`.
```
simonw: > is what lets `llm logs --json` show that

We should have that be `llm logs` as well - that should show the same information, neatly.

docs/logging.md
```diff
@@ -276,6 +285,366 @@ llm logs backup /tmp/backup.db
...
+### Logging from Python
+
+The supported way to write to a log database from Python is the `log_to_db()` method on a response, which is also what plugins should call:
+
+```python
+import llm
+import sqlite_utils
+
+db = sqlite_utils.Database("logs.db")
+model = llm.get_model("gpt-4o-mini")
```
simonw: Use "gpt-5.5" here

docs/openai-models.md
```diff
@@ -123,7 +123,7 @@ The vector size of the supported OpenAI embedding models are as follows:
 
 The `gpt-3.5-turbo-instruct` model is a little different - it is a completion model rather than a chat model, described in [the OpenAI completions documentation](https://platform.openai.com/docs/api-reference/completions/create).
 
-Completion models can be called with the `-o logprobs 3` option (not supported by chat models) which will cause LLM to store 3 log probabilities for each returned token in the SQLite database. Consult [this issue](https://github.com/simonw/llm/issues/284#issuecomment-1724772704) for details on how to read these values.
+Completion models can be called with the `-o logprobs 3` option (not supported by chat models) which will request 3 log probabilities for each returned token. These are no longer stored in the SQLite log database - raw provider payloads are not persisted there - but they can be read at runtime from `response.response_json` when using the {ref}`Python API <python-api>`. Rows logged by older versions of LLM retain the values they stored; consult [this issue](https://github.com/simonw/llm/issues/284#issuecomment-1724772704) for how to read those.
```
simonw: Let's just drop the section about `-o logprobs 3` entirely, make it an undocumented feature.

simonw:
Opus 5 and Fable 5 wrote most of this. I had GPT-5.6 Sol xhigh do a review of the diff with main:

> `This branch is ready to land - conduct an extensive code review comparing against current main branch.`
>
> `One thing to check is that this won't corrupt existing databases when it lands - it is meant to create new tables and leave existing tables alone.`

It reported:

<details><summary>three P1 issues, five P2 issues, and one P3 documentation issue</summary>

# Code review: `content-addressed-logs`

Review scope: committed branch head `84c2837` compared with `origin/main` at `5244732`. Uncommitted working-tree changes were not included in the review.

## Verdict

Not ready to land. The review found three P1 issues, five P2 issues, and one P3 documentation issue.

## Findings

### P1 — Hash attachment contents, not filesystem paths

`llm/logs.py:84`

`message.to_dict()` serializes a path-backed attachment using its path, not its contents. If the file at that path is modified in place, the message keeps the same hash. The stored attachment ID can therefore refer to the old bytes while loading the message reads the new bytes, and `verify()` can still report that the log is clean.

Resolve path-backed attachments to their bytes or another content-derived identifier before computing the message hash.

### P1 — Make the text-column migration safe to retry

`llm/migrations.py:827`

`sqlite-utils` commits each row update independently. If `m027_parts_text_column` is interrupted before the migration is recorded as complete, retrying it encounters already-migrated rows whose payload is `NULL`. Those rows are then assigned `text=None`, erasing their stored text.

Run the migration atomically and preserve or skip rows that have already been migrated.

### P1 — Restore prior conversation history for completion models

Files: `llm/cli.py:1415`, `llm/default_plugins/openai_models.py:1901`

The new loading path fills `loaded_messages`, but leaves `conversation.responses` empty. `Completion.execute()` constructs its request from `conversation.responses` plus the new prompt, so `llm -c` sends only the newest question for completion-style models.

This was reproduced with a three-message prompt chain: all three messages were loaded, but the request sent to the model contained only the final message.

Make completion execution consume the loaded prompt messages, or rehydrate the response history expected by existing completion models.

### P2 — Preserve the relationship between tool results and their parent turn

Files: `llm/logs.py:914`, `llm/logs.py:952`

`_append_turn_input()` can append a tool-result message before appending the user prompt and its attachments. In a chained turn containing tool-result attachments, a user attachment can consequently appear after the tool result. Code that relies only on `turns.parent_message_hash` and `_input_parts` can then miss the stored tool result.

The result is observable in `llm logs`: the tool result is omitted, and `-T` / `--tools` does not match it.

Store or query a direct turn relationship and an unambiguous input boundary instead of inferring ownership from part order.

### P2 — Do not treat tool-call IDs as globally unique

`llm/migrations.py:848`

`tool_instantiations` uses only `tool_call_id` as its primary key. Tool-call IDs need not be unique across independent turns, so a later call can overwrite the provenance recorded for an earlier one.

This was reproduced using the same call ID in two turns: the first log entry displayed the second turn's toolbox path.

Scope the key by turn, message, or part rather than by `tool_call_id` alone.

### P2 — Make message deduplication safe under concurrent writers

`llm/logs.py:146`

Message insertion uses a count-then-insert sequence. Two simultaneous processes can both observe that a hash is absent, after which one fails with the `UNIQUE` constraint on `messages.hash`.

This was reproduced with two SQLite connections.

Use an atomic conflict-ignore/upsert operation, and insert message parts only when the message insertion actually wins.

### P2 — Preserve repeated identical fragments

`llm/migrations.py:711`

The `turn_fragments` primary key does not include an ordering field. Passing the same fragment twice therefore collapses the two occurrences into one, regressing the duplicate-fragment support introduced by `m016`.

Include the fragment position in the key so repetitions and order are preserved.

### P2 — Resolve tool definitions in the context of each turn

`llm/logs.py:1042`

Multiple content-addressed tool definitions can have the same name. Building a global name-to-ID dictionary selects whichever definition was visited last, so an earlier turn can be reported as using the later definition.

This was reproduced with two definitions sharing a name: the first turn reported the second definition's ID.

Resolve each definition through that turn's `turn_tools` relationship.

### P3 — Complete the documented database schema

`docs/logging.md:669`

The full Cog-generated SQL table list omits `turn_search` and `tool_instantiations`, and the FTS discussion mentions only `responses_fts`.

Add the new tables and search structures to the schema documentation.

## Existing database safety

The normal upgrade from the current `main` schema to this branch was verified not to alter existing stable tables:

- A synthetic database was populated across every stable relationship, plus a custom application table.
- Before and after migration, all rows in 21 pre-existing tables and all 27 pre-existing schema objects were identical.
- `PRAGMA integrity_check` returned `ok`.
- `PRAGMA foreign_key_check` returned no violations.
- Legacy logs remained readable.

`m024` drops only the tables introduced by `m023`; when upgrading from the current stable `main`, those tables are created and replaced within the same migration sequence. It does not drop stable legacy tables.

The exception is the P1 retry problem in `m027`: databases that already contain pre-`m027` content-store rows, such as databases created from alpha or development versions, can lose migrated text if the migration is interrupted and then retried. That issue should be fixed before landing.

## Validation

For the committed review head:

- 923 tests passed with `sqlite-utils` 4.1.1.
- 923 tests passed with the minimum supported `sqlite-utils` 3.39.1.
- Black, mypy, Ruff, diff checks, and Cog checks passed.

</details>

simonw:
The sticking point on this right now involves transactions, and getting them to work in both sqlite-utils 3.x and 4.x.

I had Claude Fable (which was running out of context) write up the state of play with that problem here: https://gist.github.com/simonw/63365f6e0b1d095fec86da174c23aac7

I'm going to simplify things by bumping to sqlite-utils 4, rather than trying te keep things working across both versions. The whole point of v4 was better transactions support, time to use that.

llm/migrations.py
```diff
@@ -426,3 +425,254 @@ def m022_response_reasoning(db):
...
+  (select group_concat(json_extract(je.value, '$.literal'), '')
+     from json_each(parts.payload, '$.text_ref') je
+    where json_extract(je.value, '$.literal') is not null)
+)"""
+
+# Derives the searchable prompt and response text for turns. Serves both
+# the migration backfill (turn_filter="") and the per-turn refresh in
+# LogStore.log (turn_filter="and turns.id = :turn_id" - the slot appears
+# in three places so the filtered form touches only that turn's chain).
+TURN_SEARCH_INSERT_SQL = """
```
simonw: This should not be in the `migrations.py` file.

simonw:
Here's the 'llm logs -cue` I got back from this sequence: https://gist.github.com/simonw/c34a8c69cb3cd4d955127a4ee8672e29

```bash
llm install llm-tools-datasette
llm -T 'Datasette("https://datasette.io/content")' 'list tables'                    
llm -c -T 'Datasette("https://datasette.simonwillison.net/simonwillisonblog")' 'count blog series'
```

simonw:
Not sure about this new tool instantiations table:

CleanShot 2026-07-29 at 14 48 45@2x

simonw:
The logs now group the tools for an instance under that instance:

> # 2026-07-29T23:11:16    conversation: 01kyr2b0hzx97spnee61hf87zw id: 01kyr2b0j2h54n729y9h0fwa47
> 
> Model: **gpt-5.5**
> 
> ## Prompt
> 
> count rows in 3 interesting looking tables
> 
> ### Tools
> 
> - `Datasette({"url": "https://datasette.simonwillison.net/simonwillisonblog"})`:
>     - **Datasette_query**: `1aa6d35df9b225d03daee33201eab3ae539b67157f3a8ec544b0b5d8eaeeeeb9`  
>         Execute provided SQLite SQL query - read-only, only use SELECT  
>         Arguments: `{"sqlite_sql": {"type": "string"}}`
>     - **Datasette_schema**: `24ec61558e67e01fe6927f059ab376bad38b8dfd815851e9f3846371ef35d246`  
>         View the SQLite schema of the attached database  
>         Arguments: `{}`
> 
> ## Reasoning
> 
> **Inspecting the database schema**
> 
> I see the need for an answer that requires access to the database, so my first step is to inspect the schema. This will help ensure I have a clear understanding of the structures and data types involved. By checking the schema, I can determine how to retrieve the information effectively. It's all about getting the details right to serve the user's needs! I'll make sure I do this carefully.
> 
> ## Response
> 
> ### Tool calls
> 
> - **Datasette_schema**: `call_fBjaqeYiPCiUYzrNrB0AdLMV`  
>     Arguments: `{}`
> 
> ## Token usage
> 
> 77 input, 32 output, `{"input_tokens_details": {"cached_tokens": 0, "cache_write_tokens": 0}, "output_tokens_details": {"reasoning_tokens": 14}}`
> 
> # 2026-07-29T23:11:20
> 
> ## Prompt
> 
> -- none --
> 
> ### Tools
> 
> - `Datasette({"url": "https://datasette.simonwillison.net/simonwillisonblog"})`:
>     - **Datasette_query**: `1aa6d35`
>     - **Datasette_schema**: `24ec615`
> 
> ### Tool results
> 
> - **Datasette_schema**: `call_fBjaqeYiPCiUYzrNrB0AdLMV`  
>     ```
>     [{"group_concat(sql, ';')": "CREATE TABLE [comprehend_entity_types] (\n   [id] INTEGER PRIMARY KEY,\n   [value] TEXT\n);CREATE TABLE [comprehend_entities] (\n   [id] INTEGER PRIMARY KEY,\n   [name] TEXT,\n   [type] INTEGER REFERENCES [comprehend_entity_types]([id])\n);CREATE TABLE [blog_entry_comprehend_entities] (\n   [id] INTEGER REFERENCES [blog_entry]([id]),\n   [score] FLOAT,\n   [entity] INTEGER REFERENCES [comprehend_entities]([id]),\n   [begin_offset] INTEGER,\n   [end_offset] INTEGER\n);CREATE TABLE [blog_entry_comprehend_entities_done] (\n   [id] INTEGER PRIMARY KEY REFERENCES [blog_entry]([id])\n);CREATE TABLE [blog_entry_embeddings] (\n   [id] INTEGER PRIMARY KEY,\n   [embedding] BLOB\n);CREATE TABLE 'blog_series_fts_data'(id INTEGER PRIMARY KEY, block BLOB);CREATE TABLE 'blog_series_fts_idx'(segid, term, pgno, PRIMARY KEY(segid, term)) WITHOUT ROWID;CREATE TABLE 'blog_series_fts_docsize'(id INTEGER PRIMARY KEY, sz BLOB);CREATE TABLE 'blog_series_fts_config'(k PRIMARY KEY, v) WITHOUT ROWID;CREATE TABLE 'blog_tag_fts_data'(id INTEGER PRIMARY KEY, block BLOB);CREATE TABLE 'blog_tag_fts_idx'(segid, term, pgno, PRIMARY KEY(segid, term)) WITHOUT ROWID;CREATE TABLE 'blog_tag_fts_docsize'(id INTEGER PRIMARY KEY, sz BLOB);CREATE TABLE 'blog_tag_fts_config'(k PRIMARY KEY, v) WITHOUT ROWID;CREATE TABLE 'blog_quotation_fts_data'(id INTEGER PRIMARY KEY, block BLOB);CREATE TABLE 'blog_quotation_fts_idx'(segid, term, pgno, PRIMARY KEY(segid, term)) WITHOUT ROWID;CREATE TABLE 'blog_quotation_fts_docsize'(id INTEGER PRIMARY KEY, sz BLOB);CREATE TABLE 'blog_quotation_fts_config'(k PRIMARY KEY, v) WITHOUT ROWID;CREATE TABLE 'blog_entry_fts_data'(id INTEGER PRIMARY KEY, block BLOB);CREATE TABLE 'blog_entry_fts_idx'(segid, term, pgno, PRIMARY KEY(segid, term)) WITHOUT ROWID;CREATE TABLE 'blog_entry_fts_docsize'(id INTEGER PRIMARY KEY, sz BLOB);CREATE TABLE 'blog_entry_fts_config'(k PRIMARY KEY, v) WITHOUT ROWID;CREATE TABLE 'blog_blogmark_fts_data'(id INTEGER PRIMARY KEY, block BLOB);CREATE TABLE 'blog_blogmark_fts_idx'(segid, term, pgno, PRIMARY KEY(segid, term)) WITHOUT ROWID;CREATE TABLE 'blog_blogmark_fts_docsize'(id INTEGER PRIMARY KEY, sz BLOB);CREATE TABLE 'blog_blogmark_fts_config'(k PRIMARY KEY, v) WITHOUT ROWID;CREATE UNIQUE INDEX [idx_comprehend_entity_types_value]\n    ON [comprehend_entity_types] ([value]);CREATE UNIQUE INDEX [idx_comprehend_entities_type_name]\n    ON [comprehend_entities] ([type], [name]);CREATE VIRTUAL TABLE [blog_series_fts] USING FTS5 (\n    [title], [summary],\n    tokenize='porter',\n    content=[blog_series]\n);CREATE VIRTUAL TABLE [blog_tag_fts] USING FTS5 (\n    [tag],\n    tokenize='porter',\n    content=[blog_tag]\n);CREATE VIRTUAL TABLE [blog_quotation_fts] USING FTS5 (\n    [quotation], [source],\n    tokenize='porter',\n    content=[blog_quotation]\n);CREATE VIRTUAL TABLE [blog_entry_fts] USING FTS5 (\n    [title], [body],\n    tokenize='porter',\n    content=[blog_entry]\n);CREATE VIRTUAL TABLE [blog_blogmark_fts] USING FTS5 (\n    [link_title], [via_title], [commentary],\n    tokenize='porter',\n    content=[blog_blogmark]\n);CREATE TABLE \"blog_entry\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"title\" TEXT,\n   \"slug\" TEXT,\n   \"body\" TEXT,\n   \"created\" TEXT,\n   \"metadata\" TEXT,\n   \"tweet_html\" TEXT,\n   \"import_ref\" TEXT,\n   \"extra_head_html\" TEXT,\n   \"card_image\" TEXT,\n   \"custom_template\" TEXT,\n   \"series_id\" INTEGER REFERENCES \"blog_series\"(\"id\"),\n   \"is_draft\" INTEGER,\n   \"live_timezone\" TEXT\n);CREATE TABLE \"blog_beat_tags\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"beat_id\" INTEGER REFERENCES \"blog_beat\"(\"id\"),\n   \"tag_id\" INTEGER REFERENCES \"blog_tag\"(\"id\")\n);CREATE TABLE \"blog_previoustagname\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"previous_name\" TEXT,\n   \"tag_id\" INTEGER REFERENCES \"blog_tag\"(\"id\")\n);CREATE TABLE \"blog_blogmark\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"slug\" TEXT,\n   \"link_url\" TEXT,\n   \"link_title\" TEXT,\n   \"via_url\" TEXT,\n   \"via_title\" TEXT,\n   \"commentary\" TEXT,\n   \"created\" TEXT,\n   \"metadata\" TEXT,\n   \"import_ref\" TEXT,\n   \"card_image\" TEXT,\n   \"series_id\" TEXT REFERENCES \"blog_series\"(\"id\"),\n   \"use_markdown\" INTEGER,\n   \"is_draft\" INTEGER,\n   \"title\" TEXT\n);CREATE TABLE \"auth_permission\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"name\" TEXT,\n   \"content_type_id\" INTEGER REFERENCES \"django_content_type\"(\"id\"),\n   \"codename\" TEXT\n);CREATE TABLE \"blog_tag\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"tag\" TEXT,\n   \"description\" TEXT\n);CREATE TABLE \"blog_note\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"created\" TEXT,\n   \"slug\" TEXT,\n   \"metadata\" TEXT,\n   \"search_document\" TEXT,\n   \"import_ref\" TEXT,\n   \"card_image\" TEXT,\n   \"is_draft\" INTEGER,\n   \"body\" TEXT,\n   \"series_id\" TEXT REFERENCES \"blog_series\"(\"id\"),\n   \"title\" TEXT\n);CREATE TABLE \"guides_guide\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"created\" TEXT,\n   \"updated\" TEXT,\n   \"title\" TEXT,\n   \"slug\" TEXT,\n   \"description\" TEXT,\n   \"is_draft\" INTEGER\n);CREATE TABLE \"blog_quotation_tags\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"quotation_id\" INTEGER REFERENCES \"blog_quotation\"(\"id\"),\n   \"tag_id\" INTEGER REFERENCES \"blog_tag\"(\"id\")\n);CREATE TABLE \"blog_beat\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"created\" TEXT,\n   \"slug\" TEXT,\n   \"metadata\" TEXT,\n   \"search_document\" TEXT,\n   \"import_ref\" TEXT,\n   \"card_image\" TEXT,\n   \"is_draft\" INTEGER,\n   \"beat_type\" TEXT,\n   \"title\" TEXT,\n   \"url\" TEXT,\n   \"commentary\" TEXT,\n   \"series_id\" TEXT REFERENCES \"blog_series\"(\"id\"),\n   \"image_alt\" TEXT,\n   \"image_url\" TEXT,\n   \"note\" TEXT\n);CREATE TABLE \"redirects_redirect\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"domain\" TEXT,\n   \"path\" TEXT,\n   \"target\" TEXT,\n   \"created\" TEXT\n);CREATE TABLE \"blog_note_tags\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"note_id\" INTEGER REFERENCES \"blog_note\"(\"id\"),\n   \"tag_id\" INTEGER REFERENCES \"blog_tag\"(\"id\")\n);CREATE TABLE \"blog_entry_tags\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"entry_id\" INTEGER REFERENCES \"blog_entry\"(\"id\"),\n   \"tag_id\" INTEGER REFERENCES \"blog_tag\"(\"id\")\n);CREATE TABLE \"guides_chapter\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"created\" TEXT,\n   \"slug\" TEXT,\n   \"metadata\" TEXT,\n   \"search_document\" TEXT,\n   \"import_ref\" TEXT,\n   \"card_image\" TEXT,\n   \"is_draft\" INTEGER,\n   \"updated\" TEXT,\n   \"title\" TEXT,\n   \"body\" TEXT,\n   \"order\" INTEGER,\n   \"series_id\" TEXT REFERENCES \"blog_series\"(\"id\"),\n   \"guide_id\" INTEGER REFERENCES \"guides_guide\"(\"id\"),\n   \"section_id\" INTEGER REFERENCES \"guides_guidesection\"(\"id\"),\n   \"is_unlisted\" INTEGER\n);CREATE TABLE \"guides_guidesection\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"title\" TEXT,\n   \"slug\" TEXT,\n   \"order\" INTEGER,\n   \"guide_id\" INTEGER REFERENCES \"guides_guide\"(\"id\")\n);CREATE TABLE \"blog_quotation\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"slug\" TEXT,\n   \"quotation\" TEXT,\n   \"source\" TEXT,\n   \"source_url\" TEXT,\n   \"created\" TEXT,\n   \"metadata\" TEXT,\n   \"import_ref\" TEXT,\n   \"card_image\" TEXT,\n   \"series_id\" TEXT REFERENCES \"blog_series\"(\"id\"),\n   \"is_draft\" INTEGER,\n   \"context\" TEXT\n);CREATE TABLE \"auth_user\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"password\" TEXT,\n   \"last_login\" TEXT,\n   \"is_superuser\" INTEGER,\n   \"username\" TEXT,\n   \"first_name\" TEXT,\n   \"last_name\" TEXT,\n   \"email\" TEXT,\n   \"is_staff\" INTEGER,\n   \"is_active\" INTEGER,\n   \"date_joined\" TEXT\n);CREATE TABLE \"blog_blogmark_tags\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"blogmark_id\" INTEGER REFERENCES \"blog_blogmark\"(\"id\"),\n   \"tag_id\" INTEGER REFERENCES \"blog_tag\"(\"id\")\n);CREATE TABLE \"blog_series\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"created\" TEXT,\n   \"slug\" TEXT,\n   \"title\" TEXT,\n   \"summary\" TEXT\n);CREATE TABLE \"django_content_type\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"app_label\" TEXT,\n   \"model\" TEXT\n);CREATE TABLE til(\n  path TEXT,\n  topic TEXT,\n  title TEXT,\n  url TEXT,\n  body TEXT,\n  html TEXT,\n  created TEXT,\n  created_utc TEXT,\n  updated TEXT,\n  updated_utc TEXT,\n  shot_hash TEXT,\n  slug TEXT,\n  summary TEXT\n)"}]
>     ```
> 
> ## Reasoning
> 
> **Considering row counts in tables**
> 
> I see that the user's request to "count rows in 3 interesting looking tables" is ambiguous. I need to define what “interesting” means in this context before I proceed. I think I could choose tables like blog_entry and blog_blogmark, among others, to count. So, it looks like I’ll need to create a query to get those row counts. Let’s clarify which tables are considered interesting!
> 
> ## Response
> 
> ### Tool calls
> 
> - **Datasette_query**: `call_hTvC5fExICd7lKgz8hiFZTfC`  
>     sqlite_sql:
>     ```
>     SELECT 'blog_entry' AS table_name, COUNT(*) AS row_count FROM blog_entry
>     UNION ALL
>     SELECT 'blog_blogmark' AS table_name, COUNT(*) AS row_count FROM blog_blogmark
>     UNION ALL
>     SELECT 'til' AS table_name, COUNT(*) AS row_count FROM til
>     ORDER BY table_name;
>     ```
> 
> ## Token usage
> 
> 2,614 input, 127 output, `{"input_tokens_details": {"cached_tokens": 0, "cache_write_tokens": 0}, "output_tokens_details": {"reasoning_tokens": 38}}`
> 
> # 2026-07-29T23:11:25
> 
> ## Prompt
> 
> -- none --
> 
> ### Tools
> 
> - `Datasette({"url": "https://datasette.simonwillison.net/simonwillisonblog"})`:
>     - **Datasette_query**: `1aa6d35`
>     - **Datasette_schema**: `24ec615`
> 
> ### Tool results
> 
> - **Datasette_query**: `call_hTvC5fExICd7lKgz8hiFZTfC`  
>     ```
>     [{"table_name": "blog_blogmark", "row_count": 8422}, {"table_name": "blog_entry", "row_count": 3318}, {"table_name": "til", "row_count": 581}]
>     ```
> 
> ## Response
> 
> Counts for three interesting-looking tables:
> 
> | table | rows |
> |---|---:|
> | `blog_blogmark` | 8,422 |
> | `blog_entry` | 3,318 |
> | `til` | 581 |
> 
> ## Token usage
> 
> 2,802 input, 53 output, `{"input_tokens_details": {"cached_tokens": 2048, "cache_write_tokens": 0}, "output_tokens_details": {"reasoning_tokens": 0}}`

simonw:
I had GPT-5.5 create a new description of this PR just using the diff:
```bash
llm -f https://patch-diff.githubusercontent.com/raw/simonw/llm/pull/1562.diff -s 'explain this change in detail, make it nice to read'
```
It's really good:

<details><summary>GPT-5.5 explanation of the diff</summary>

## Overview

This is a major overhaul of LLM’s logging system. The old logging model stored each request/response as a flat row in the legacy `responses` table. This change introduces a new **content-addressed message store** built around conversations as chains of messages and parts.

The short version:

- New logs are written to new tables: `threads`, `turns`, `messages`, `parts`, and related tables.
- Old `responses` rows are no longer written, but they are still read.
- `llm logs` now merges old and new log data into one view.
- Message content is deduplicated by hash, so shared history is stored once.
- Logging writes now use real transactions via `sqlite-utils>=4.0`.
- Full-text search now works across both old and new logs, with better ranking.
- Conversation continuation now works from the new message store.
- The docs and tests have been substantially updated to describe and verify the new model.

---

## The big architectural change: a content-addressed message store

The largest addition is the new `llm/logs.py` module. It introduces `LogStore`, which is now responsible for reading and writing conversation history.

Previously, logging was centered on this shape:

```text
responses
  id
  prompt
  response
  prompt_json
  response_json
  conversation_id
  ...
```

That was simple, but it flattened conversations into rows. It also duplicated content, made forks and shared history hard to represent, and relied on raw provider payloads for some information.

The new design stores conversations as a tree of messages:

```text
threads
  conversations

turns
  individual model calls

messages
  content-addressed message nodes

parts
  ordered pieces of a message:
  text, reasoning, tool calls, tool results, attachments
```

A **message** is identified by a hash of:

1. Its canonical content.
2. Its parent message hash.

That means the same message content in a different conversation position gets a different hash, but two conversations that share the same prefix can share the same stored rows.

For example, if two conversations begin the same way and later diverge, the shared beginning is stored only once.

---

## New hashing contract

`llm/logs.py` adds:

```python
canonical_json()
content_hash()
message_hash()
```

Hashes use:

- canonical JSON
- sorted keys
- compact separators
- UTF-8
- BLAKE2b
- 16-byte digest
- `b2:` prefix

So hashes look like:

```text
b2:d6b0cd4e7a65ea90423c50fadb3f5704
```

The `b2:` prefix is intentional: it marks the hashing algorithm so future changes can be detected rather than silently mixing incompatible hash formats.

The hash covers the **resolved content** of a message. That matters for fragments and attachments:

- Fragment references are expanded when hashing.
- Attachment identity is based on content plus media type.
- Path-backed attachments are re-read during verification.
- URL attachments are hashed by URL.

This makes the hash a contract for “what the model actually saw”.

---

## New database tables

The migrations add a set of new tables:

```text
messages
parts
part_attachments
part_fragments
threads
turns
turn_tools
turn_fragments
turn_search
turn_search_fts
tool_instantiations
```

### `messages`

Stores one unique message node.

Important columns:

```text
hash
parent_hash
role
provider_metadata
```

The parent hash links messages into a chain.

### `parts`

Stores the contents of a message in order.

Important columns:

```text
message_hash
position
type
tool_name
text
payload
```

A notable storage improvement: literal text now lives in a dedicated `text` column instead of being wrapped in JSON. That means ordinary prose is stored as ordinary SQL text, not escaped JSON.

The `payload` column is now reserved for structured data such as:

- fragment references
- attachment references
- tool call fields
- provider metadata

The redundant `type` key is removed from the JSON payload because `type` is now its own column.

### `threads`

Represents conversations.

A thread points at the current tip of a message chain:

```text
id
name
tip_message_hash
forked_from
datetime_utc
```

Thread IDs are conversation IDs.

### `turns`

Represents one model call.

A turn records:

```text
id
thread_id
parent_message_hash
tip_message_hash
model
resolved_model
options_json
schema_id
input_tokens
output_tokens
token_details
duration_ms
datetime_utc
```

The pair of hashes:

```text
parent_message_hash
tip_message_hash
```

describes exactly what the turn added to the chain.

### `turn_fragments`

Tracks which fragments were supplied to a specific model call.

This is separate from `part_fragments` because message rows are shared, but fragment usage is turn-specific provenance.

### `turn_search` and `turn_search_fts`

These power full-text search for the new log tables.

Searchable text is derived from:

- the typed user prompt
- the model’s text response

It deliberately excludes:

- system prompts
- fragment contents
- tool calls
- tool results
- reasoning traces

---

## Logging no longer writes the legacy tables

`Response.log_to_db()` has been completely rewritten.

Previously it inserted into:

```text
conversations
responses
prompt_fragments
system_fragments
prompt_attachments
tool_responses
tool_calls
tool_results
...
```

Now it delegates to:

```python
LogStore(db).log(self)
```

New logging writes only the content-addressed message store.

The legacy tables remain in place, but they are now read-only history. This is an important compatibility choice: existing logs are not discarded, but new logs use the new schema.

---

## Raw provider payloads are no longer persisted

The old logger stored:

```text
prompt_json
response_json
```

Those columns often contained condensed raw provider request and response payloads.

The new logging system does not persist those raw provider payloads. Instead, the message chain is the durable record of what was sent and returned.

This has one visible consequence called out in the changelog:

- OpenAI completion `logprobs` data is no longer stored in the database.
- It remains available on the response object through the Python API.
- Old rows that already recorded it keep their data.

The OpenAI docs were updated to remove the previous claim that completion `logprobs` are stored in SQLite.

---

## `llm logs` now merges old and new logs

The CLI’s log-reading code has been refactored heavily.

Old SQL constants and helper queries were removed from `llm/cli.py`. Instead, the CLI now imports from `llm.logs`:

```python
LogStore
merged_log_rows
log_row_extras
legacy_log_row_extras
```

`merged_log_rows()` reads from both:

- new `turns` / message-store tables
- old legacy `responses` table

It deduplicates by ID, so rows from a dual-write or migration era do not appear twice.

This means after upgrading, users still see their old logs:

```bash
llm logs
```

but new rows come from the content-addressed store.

The filters also work across both generations:

- `-m`
- `-c`
- `-f`
- `-T`
- `--tools`
- `--schema`
- ID filters
- search queries

---

## Conversation continuation now uses threads

Conversation loading has been updated to look at both old and new storage.

Previously, `-c` looked for the latest row in `conversations`.

Now it checks both:

```sql
select id from threads
union
select id from conversations
```

If a legacy conversation exists, it can still be reconstructed from `responses`.

If a new thread exists, the exact message chain is loaded from `LogStore.thread_messages()` and attached to the conversation as:

```python
conversation.loaded_messages
```

That matters because the new store preserves richer structure than the old flat rows could, including:

- reasoning parts
- provider metadata
- tool messages
- attachment parts

When a continued conversation sends a new prompt, it now uses that stored chain as its history.

---

## Responses outside conversations now still get threads

Previously, the legacy logging path would create a conversation record even for a one-off logged response. The new logger preserves that guarantee by creating a thread for responses logged outside an explicit conversation.

This matters for the Python API:

```python
response = model.prompt("Hello")
response.text()
response.log_to_db(db)
```

That logged response can now still be found and continued with:

```bash
llm -c
```

---

## `messages=` behavior was fixed

There is an important behavioral fix around calls like:

```python
model.prompt(
    "follow-up question",
    messages=[llm.user("previous message")]
)
```

Previously, if `messages=` was supplied, the explicit messages list could be treated as the whole chain, and the separate `prompt=` text could be omitted from `prompt.messages`.

That meant the model might receive text that was not represented in the logged message chain.

This change introduces `_append_turn_input()` and updates model/conversation prompt construction so that:

- `messages=` is the authoritative existing history
- `prompt=`, `fragments=`, `attachments=`, and `tool_results=` are appended as the new turn input

So this now produces a chain like:

```python
[
    llm.user("previous message"),
    llm.user("follow-up question"),
]
```

instead of silently dropping the follow-up from the chain.

This fix applies to:

- sync model prompts
- async model prompts
- conversations
- chains

---

## Fragments are stored by reference

Fragments are now represented more efficiently in message parts.

If a prompt includes a large fragment, the resolved prompt text is still what gets hashed, but the stored part can reference the fragment instead of duplicating the text.

For example, instead of storing a giant string repeatedly, the part payload can hold:

```json
{
  "text_ref": [
    {"fragment": 1},
    {"literal": "\nquestion about it"}
  ]
}
```

This means asking many questions about the same large document stores the document once.

The new `part_fragments` table makes those references queryable with normal joins.

---

## Attachments are stored by reference too

Attachments continue to use the existing content-addressed `attachments` table.

The new part payloads reference attachment IDs instead of embedding attachment data directly.

There are also new junction tables:

```text
part_attachments
tool_results_attachments
```

Attachment hashing is careful about identity:

- content bytes matter
- media type matters
- filesystem path itself does not define identity
- path-backed files are re-read during verification

So the same bytes at two different paths are treated as the same attachment identity, but the same bytes with different media types are treated as different model inputs.

---

## `LogStore.verify()`

A new verification feature can re-derive every stored message hash from the database contents:

```python
LogStore(db).verify()
```

It returns a list of broken message hashes.

This catches cases where stored references no longer resolve to the original content, such as:

- corrupted part text
- deleted fragment rows
- changed path-backed attachment files
- missing path-backed attachment files

An empty list means the stored chains still match their hashes.

---

## Full-text search was rewritten and improved

`llm logs -q` now searches both generations of logs.

For new logs, it uses:

```text
turn_search
turn_search_fts
```

For old logs, it still uses:

```text
responses_fts
```

Search covers only:

- prompt text typed by the user
- assistant text responses

It excludes:

- system prompts
- fragment content
- tool activity
- reasoning traces

Search ranking now uses SQLite FTS5 BM25 scoring with prompt matches weighted more strongly than response matches:

```python
bm25(turn_search_fts, 10.0, 1.0)
```

A longstanding ranking bug was also fixed. SQLite BM25 scores are “lower is better”, and the old implementation sorted in the wrong direction, so weaker matches could appear first. This change sorts best matches first.

`-l/--latest` can still switch search results to most-recent-first.

Invalid FTS syntax is now handled with a clean CLI error that points users to the SQLite FTS5 query syntax documentation.

---

## Tool logging now records configured toolbox instances

Tool logging has been improved so `llm logs` can show which configured toolbox instance supplied a tool.

For example, a tool result can now indicate that:

```text
SQLite_query
```

came from:

```text
SQLite("mydb.db")
```

This is handled through:

```text
tool_instances
tool_instantiations
turn_tools.instance_id
```

Each distinct toolbox configuration is stored once in `tool_instances`.

Then:

- `turn_tools` records which configured instance made a tool available to a turn.
- `tool_instantiations` records which configured instance served a specific tool call.

The CLI rendering was updated so tools from the same instance are grouped visually beneath the instance line.

---

## Chain responses now insert display-only spaces between rounds

Chain responses can run multiple model calls in sequence. Previously, streamed text from two rounds could be concatenated directly:

```text
...end of first responseStart of second response...
```

This change inserts a single space at the boundary when neither side already supplies whitespace.

The key detail: the separator is synthesized at the chain level for display only. It is not recorded as a real response event and therefore is not logged as a whitespace-only message part.

This avoids both ugly output and polluted logs.

---

## Async logging preserves stream events better

`AsyncResponse.to_sync_response()` now copies `_stream_events`.

That matters because the CLI converts async responses into sync responses before logging. Without copying stream events, structured parts such as reasoning or provider metadata could be lost and reduced to plain text.

This fix preserves richer async response structure in the message store.

---

## OpenAI completion continuation now uses `prompt.messages`

The OpenAI completion model implementation was updated.

Previously it reconstructed history from:

```python
conversation.responses
```

That does not work for conversations reloaded from the new message store, because their history lives in `prompt.messages` / `loaded_messages`.

Now completion models build their prompt from the full message chain:

```python
for message in prompt.messages:
    ...
```

This fixes continuation for OpenAI completion models after loading a conversation from storage.

---

## Transactions now use `sqlite-utils` 4.0 `Database.atomic()`

The dependency changed from:

```toml
sqlite-utils>=3.37
```

to:

```toml
sqlite-utils>=4.0
```

The reason is important: writes now use real transactions with:

```python
with db.atomic():
    ...
```

Several old uses of:

```python
with db.conn:
    ...
```

were removed or replaced.

This affects:

- logging turns
- writing message chains
- storing fragments and aliases
- embedding batches
- deleting embedding collections
- embedding migrations
- fragment commands

The goal is to avoid partial writes. For example, if logging a message writes the `messages` row but crashes before writing `parts`, the database would be left in a corrupt state. Real transactions allow the whole write to roll back cleanly.

---

## Migrations added

Three new migrations are introduced.

### `m023_message_store`

Creates the main content-addressed message store tables:

```text
messages
parts
part_attachments
part_fragments
threads
turns
turn_tools
turn_fragments
turn_search
turn_search_fts
tool_instantiations
```

### `m024_tool_instance_references`

Normalizes tool instance configuration storage.

It changes `tool_instantiations` to reference `tool_instances` by `instance_id` instead of copying instance fields directly.

### `m025_turn_tools_instance_backfill`

Backfills `turn_tools.instance_id` for development-era rows that were written before that column existed.

---

## Documentation updates

The changelog gets a large new “Unreleased” section explaining the user-facing implications.

The logging documentation gains a substantial new section:

```text
The message store
```

It explains:

- threads, turns, messages, and parts
- content addressing
- the hash contract
- shared history
- forking
- storage by reference
- table-by-table schema notes
- example SQL queries
- logging from Python

The generated SQL schema docs were also updated to include the new tables.

The README table of contents was regenerated to include the new logging documentation sections.

The OpenAI models docs were updated to remove the note claiming `logprobs` are stored in SQLite.

---

## Test suite updates

A large new test file was added:

```text
tests/test_logs_store.py
```

It covers the new message store extensively, including:

- canonical JSON behavior
- message hash stability
- schema creation
- round-tripping message chains
- deduplication
- thread behavior
- forking
- pending tool calls
- response logging
- CLI integration
- continuation from stored messages
- legacy fallback
- fragment references
- attachment hashing
- verification
- async logging
- `llm logs` behavior against new tables
- tool instance provenance
- transaction rollback behavior
- concurrent writer behavior

Many existing tests were updated from expecting rows in `responses` to expecting rows in `turns` or reading through `merged_log_rows()`.

Search tests were expanded to cover:

- new-table search
- prompt vs response ranking
- excluding fragment content
- merging legacy rows
- invalid FTS queries

Tool tests were updated because tool calls and results now live in message parts rather than the legacy `tool_calls` / `tool_results` tables.

---

## CI and tooling changes

The GitHub Actions matrix was simplified.

Previously CI explicitly tested with both:

```text
sqlite-utils<4
sqlite-utils>=4
```

That is no longer needed because the package now requires:

```text
sqlite-utils>=4.0
```

The special `sqlite-utils==4.0rc2` override was also removed from the `Justfile`.

Cog checks now just run with the normal dependency set.

`.gitignore` also now ignores:

```text
backups-do-not-delete/
```

---

## User-facing impact

For most users, the main visible behavior is:

- `llm logs` still shows old logs.
- New logs are stored in a richer schema.
- `llm logs -q` search is better and works across old and new logs.
- Conversation continuation works from the new message store.
- Tool logs show configured toolbox instance information.
- Logs no longer include raw provider payloads.
- OpenAI completion `logprobs` are no longer persisted to SQLite.
- Database writes are safer because they are transactional.

For developers and plugin authors:

- `response.log_to_db(db)` remains the supported logging API.
- It now writes to the content-addressed store.
- `LogStore` exists internally to manage the new schema and behavior.
- The documented table schema is now the supported database interface for querying logs directly.

</details>

simonw:
I'm trying to confirm that storage of forked message threads works, by firing up https://github.com/simonw/llm-chat-completions-server against the latest version and running forked chat completions API calls against it.

Doesn't look like it worked, each one shows as a separate thread:

Image

That's using this (Claude Fable written) SQL query:

<details><summary>SQL query</summary>

```sql
with per_thread as (
  select thread_id, count(*) as turns,
    sum(coalesce(input_tokens, 0) + coalesce(output_tokens, 0)) as tokens,
    group_concat(distinct model) as models,
    max(datetime_utc) as last_turn
  from turns where thread_id is not null group by thread_id
),
chain as (
  -- walk each thread's message chain from tip back to root
  select th.id as thread_id, m.hash, m.parent_hash
  from threads th join messages m on m.hash = th.tip_message_hash
  union all
  select c.thread_id, m.hash, m.parent_hash
  from chain c join messages m on m.hash = c.parent_hash
)
select
  case when th.forked_from is not null then '⑂ ' else '' end ||
    substr(coalesce(
      (select ts.prompt from turn_search ts
       join turns t on t.id = ts.turn_id
       where t.thread_id = th.id
       order by t.datetime_utc limit 1),
      th.name), 1, 44) as first_prompt,
  pt.turns,
  (select count(*) from chain c where c.thread_id = th.id) as msgs,
  strftime('%Y-%m-%d %H:%M:%S', th.datetime_utc) as started,
  pt.models,
  (select json_group_array(name) from (
     select distinct p.tool_name as name
     from chain c join parts p on p.message_hash = c.hash
     where c.thread_id = th.id and p.type = 'tool_result'
       and p.tool_name is not null
     order by name
  )) as tools,
  pt.tokens
from threads th
left join per_thread pt on pt.thread_id = th.id
order by coalesce(pt.last_turn, th.datetime_utc) desc
limit 12;
```
</details>

simonw:
Here's a tree!

CleanShot 2026-07-29 at 17 05 28@2x

<details><summary>SQL query</summary>

```sql
with msg as (
  select m.hash, m.parent_hash, m.role, m.rowid as rid,
    replace(coalesce(
      nullif(p.text, ''),
      (select f.content from part_fragments pf
       join fragments f on f.id = pf.fragment_id
       where pf.part_id = p.id
       order by pf."order" limit 1),
      '[' || coalesce(p.type, 'empty') || ']'
    ), char(10), ' ') as text,
    (select group_concat(p2.tool_name, ', ') from parts p2
     where p2.message_hash = m.hash and p2.type = 'tool_result'
       and p2.tool_name is not null) as tools
  from messages m
  left join parts p on p.message_hash = m.hash and p.position = 0
),
tree as (
  select hash, role, text, tools, 0 as depth,
    printf('%012d', rid) as path, hash as conversation
  from msg where parent_hash is null
  union all
  select msg.hash, msg.role, msg.text, msg.tools, t.depth + 1,
    t.path || '/' || printf('%012d', msg.rid),
    t.conversation
  from msg join tree t on msg.parent_hash = t.hash
),
turn_chain as (
  select t.id as turn_id, t.datetime_utc, m.hash, m.parent_hash
  from turns t join messages m on m.hash = t.tip_message_hash
  union all
  select tc.turn_id, tc.datetime_utc, m.hash, m.parent_hash
  from turn_chain tc join messages m on m.hash = tc.parent_hash
)
select
  t.conversation,
  strftime('%Y-%m-%d %H:%M:%S',
    (select min(tc.datetime_utc) from turn_chain tc where tc.hash = t.hash)
  ) as datetime,
  replace(hex(zeroblob(t.depth)), '00', '    ') || substr(t.text, 1, 60) as message,
  coalesce(t.tools, '') as tools
from tree t
order by t.path;
```
</details>

simonw:
I'm adding that tree query as a view, then we can facet by it easily.

I'm also going to try and get `llm -c` working against conversations with a configured toolbox instance:
- #1092

simonw:
This worked!
```bash
uv run --with llm-tools-datasette --with-editable . llm -T 'Datasette("https://datasette.simonwillison.net/simonwillisonblog")' 'count entries'
uv run --with llm-tools-datasette --with-editable . llm -c 'now blogmarks'​
```

<details><summary>llm logs -cue</summary>

# 2026-07-30T00:13:50    conversation: 01kyr5xjwr7jfmkc22tb7t0p2w id: 01kyr5xjwve0x8rbmqacttgwz7

Model: **gpt-5.5**

## Prompt

count entries

### Tools

- `Datasette({"url": "https://datasette.simonwillison.net/simonwillisonblog"})`:
    - **Datasette_query**: `1aa6d35df9b225d03daee33201eab3ae539b67157f3a8ec544b0b5d8eaeeeeb9`  
        Execute provided SQLite SQL query - read-only, only use SELECT  
        Arguments: `{"sqlite_sql": {"type": "string"}}`
    - **Datasette_schema**: `24ec61558e67e01fe6927f059ab376bad38b8dfd815851e9f3846371ef35d246`  
        View the SQLite schema of the attached database  
        Arguments: `{}`

## Response

### Tool calls

- **Datasette_schema**: `call_tuQSDIhoHVhrMzWiv54wyf9K`  
    Arguments: `{}`

## Token usage

71 input, 34 output, `{"input_tokens_details": {"cache_write_tokens": 0, "cached_tokens": 0}, "output_tokens_details": {"reasoning_tokens": 16}}`

# 2026-07-30T00:13:54

## Prompt

-- none --

### Tools

- `Datasette({"url": "https://datasette.simonwillison.net/simonwillisonblog"})`:
    - **Datasette_query**: `1aa6d35`
    - **Datasette_schema**: `24ec615`

### Tool results

- **Datasette_schema**: `call_tuQSDIhoHVhrMzWiv54wyf9K`  
    ```
    [{"group_concat(sql, ';')": "CREATE TABLE [comprehend_entity_types] (\n   [id] INTEGER PRIMARY KEY,\n   [value] TEXT\n);CREATE TABLE [comprehend_entities] (\n   [id] INTEGER PRIMARY KEY,\n   [name] TEXT,\n   [type] INTEGER REFERENCES [comprehend_entity_types]([id])\n);CREATE TABLE [blog_entry_comprehend_entities] (\n   [id] INTEGER REFERENCES [blog_entry]([id]),\n   [score] FLOAT,\n   [entity] INTEGER REFERENCES [comprehend_entities]([id]),\n   [begin_offset] INTEGER,\n   [end_offset] INTEGER\n);CREATE TABLE [blog_entry_comprehend_entities_done] (\n   [id] INTEGER PRIMARY KEY REFERENCES [blog_entry]([id])\n);CREATE TABLE [blog_entry_embeddings] (\n   [id] INTEGER PRIMARY KEY,\n   [embedding] BLOB\n);CREATE TABLE 'blog_series_fts_data'(id INTEGER PRIMARY KEY, block BLOB);CREATE TABLE 'blog_series_fts_idx'(segid, term, pgno, PRIMARY KEY(segid, term)) WITHOUT ROWID;CREATE TABLE 'blog_series_fts_docsize'(id INTEGER PRIMARY KEY, sz BLOB);CREATE TABLE 'blog_series_fts_config'(k PRIMARY KEY, v) WITHOUT ROWID;CREATE TABLE 'blog_tag_fts_data'(id INTEGER PRIMARY KEY, block BLOB);CREATE TABLE 'blog_tag_fts_idx'(segid, term, pgno, PRIMARY KEY(segid, term)) WITHOUT ROWID;CREATE TABLE 'blog_tag_fts_docsize'(id INTEGER PRIMARY KEY, sz BLOB);CREATE TABLE 'blog_tag_fts_config'(k PRIMARY KEY, v) WITHOUT ROWID;CREATE TABLE 'blog_quotation_fts_data'(id INTEGER PRIMARY KEY, block BLOB);CREATE TABLE 'blog_quotation_fts_idx'(segid, term, pgno, PRIMARY KEY(segid, term)) WITHOUT ROWID;CREATE TABLE 'blog_quotation_fts_docsize'(id INTEGER PRIMARY KEY, sz BLOB);CREATE TABLE 'blog_quotation_fts_config'(k PRIMARY KEY, v) WITHOUT ROWID;CREATE TABLE 'blog_entry_fts_data'(id INTEGER PRIMARY KEY, block BLOB);CREATE TABLE 'blog_entry_fts_idx'(segid, term, pgno, PRIMARY KEY(segid, term)) WITHOUT ROWID;CREATE TABLE 'blog_entry_fts_docsize'(id INTEGER PRIMARY KEY, sz BLOB);CREATE TABLE 'blog_entry_fts_config'(k PRIMARY KEY, v) WITHOUT ROWID;CREATE TABLE 'blog_blogmark_fts_data'(id INTEGER PRIMARY KEY, block BLOB);CREATE TABLE 'blog_blogmark_fts_idx'(segid, term, pgno, PRIMARY KEY(segid, term)) WITHOUT ROWID;CREATE TABLE 'blog_blogmark_fts_docsize'(id INTEGER PRIMARY KEY, sz BLOB);CREATE TABLE 'blog_blogmark_fts_config'(k PRIMARY KEY, v) WITHOUT ROWID;CREATE UNIQUE INDEX [idx_comprehend_entity_types_value]\n    ON [comprehend_entity_types] ([value]);CREATE UNIQUE INDEX [idx_comprehend_entities_type_name]\n    ON [comprehend_entities] ([type], [name]);CREATE VIRTUAL TABLE [blog_series_fts] USING FTS5 (\n    [title], [summary],\n    tokenize='porter',\n    content=[blog_series]\n);CREATE VIRTUAL TABLE [blog_tag_fts] USING FTS5 (\n    [tag],\n    tokenize='porter',\n    content=[blog_tag]\n);CREATE VIRTUAL TABLE [blog_quotation_fts] USING FTS5 (\n    [quotation], [source],\n    tokenize='porter',\n    content=[blog_quotation]\n);CREATE VIRTUAL TABLE [blog_entry_fts] USING FTS5 (\n    [title], [body],\n    tokenize='porter',\n    content=[blog_entry]\n);CREATE VIRTUAL TABLE [blog_blogmark_fts] USING FTS5 (\n    [link_title], [via_title], [commentary],\n    tokenize='porter',\n    content=[blog_blogmark]\n);CREATE TABLE \"blog_entry\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"title\" TEXT,\n   \"slug\" TEXT,\n   \"body\" TEXT,\n   \"created\" TEXT,\n   \"metadata\" TEXT,\n   \"tweet_html\" TEXT,\n   \"import_ref\" TEXT,\n   \"extra_head_html\" TEXT,\n   \"card_image\" TEXT,\n   \"custom_template\" TEXT,\n   \"series_id\" INTEGER REFERENCES \"blog_series\"(\"id\"),\n   \"is_draft\" INTEGER,\n   \"live_timezone\" TEXT\n);CREATE TABLE \"blog_beat_tags\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"beat_id\" INTEGER REFERENCES \"blog_beat\"(\"id\"),\n   \"tag_id\" INTEGER REFERENCES \"blog_tag\"(\"id\")\n);CREATE TABLE \"blog_previoustagname\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"previous_name\" TEXT,\n   \"tag_id\" INTEGER REFERENCES \"blog_tag\"(\"id\")\n);CREATE TABLE \"blog_blogmark\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"slug\" TEXT,\n   \"link_url\" TEXT,\n   \"link_title\" TEXT,\n   \"via_url\" TEXT,\n   \"via_title\" TEXT,\n   \"commentary\" TEXT,\n   \"created\" TEXT,\n   \"metadata\" TEXT,\n   \"import_ref\" TEXT,\n   \"card_image\" TEXT,\n   \"series_id\" TEXT REFERENCES \"blog_series\"(\"id\"),\n   \"use_markdown\" INTEGER,\n   \"is_draft\" INTEGER,\n   \"title\" TEXT\n);CREATE TABLE \"auth_permission\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"name\" TEXT,\n   \"content_type_id\" INTEGER REFERENCES \"django_content_type\"(\"id\"),\n   \"codename\" TEXT\n);CREATE TABLE \"blog_tag\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"tag\" TEXT,\n   \"description\" TEXT\n);CREATE TABLE \"blog_note\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"created\" TEXT,\n   \"slug\" TEXT,\n   \"metadata\" TEXT,\n   \"search_document\" TEXT,\n   \"import_ref\" TEXT,\n   \"card_image\" TEXT,\n   \"is_draft\" INTEGER,\n   \"body\" TEXT,\n   \"series_id\" TEXT REFERENCES \"blog_series\"(\"id\"),\n   \"title\" TEXT\n);CREATE TABLE \"guides_guide\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"created\" TEXT,\n   \"updated\" TEXT,\n   \"title\" TEXT,\n   \"slug\" TEXT,\n   \"description\" TEXT,\n   \"is_draft\" INTEGER\n);CREATE TABLE \"blog_quotation_tags\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"quotation_id\" INTEGER REFERENCES \"blog_quotation\"(\"id\"),\n   \"tag_id\" INTEGER REFERENCES \"blog_tag\"(\"id\")\n);CREATE TABLE \"blog_beat\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"created\" TEXT,\n   \"slug\" TEXT,\n   \"metadata\" TEXT,\n   \"search_document\" TEXT,\n   \"import_ref\" TEXT,\n   \"card_image\" TEXT,\n   \"is_draft\" INTEGER,\n   \"beat_type\" TEXT,\n   \"title\" TEXT,\n   \"url\" TEXT,\n   \"commentary\" TEXT,\n   \"series_id\" TEXT REFERENCES \"blog_series\"(\"id\"),\n   \"image_alt\" TEXT,\n   \"image_url\" TEXT,\n   \"note\" TEXT\n);CREATE TABLE \"redirects_redirect\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"domain\" TEXT,\n   \"path\" TEXT,\n   \"target\" TEXT,\n   \"created\" TEXT\n);CREATE TABLE \"blog_note_tags\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"note_id\" INTEGER REFERENCES \"blog_note\"(\"id\"),\n   \"tag_id\" INTEGER REFERENCES \"blog_tag\"(\"id\")\n);CREATE TABLE \"blog_entry_tags\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"entry_id\" INTEGER REFERENCES \"blog_entry\"(\"id\"),\n   \"tag_id\" INTEGER REFERENCES \"blog_tag\"(\"id\")\n);CREATE TABLE \"guides_chapter\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"created\" TEXT,\n   \"slug\" TEXT,\n   \"metadata\" TEXT,\n   \"search_document\" TEXT,\n   \"import_ref\" TEXT,\n   \"card_image\" TEXT,\n   \"is_draft\" INTEGER,\n   \"updated\" TEXT,\n   \"title\" TEXT,\n   \"body\" TEXT,\n   \"order\" INTEGER,\n   \"series_id\" TEXT REFERENCES \"blog_series\"(\"id\"),\n   \"guide_id\" INTEGER REFERENCES \"guides_guide\"(\"id\"),\n   \"section_id\" INTEGER REFERENCES \"guides_guidesection\"(\"id\"),\n   \"is_unlisted\" INTEGER\n);CREATE TABLE \"guides_guidesection\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"title\" TEXT,\n   \"slug\" TEXT,\n   \"order\" INTEGER,\n   \"guide_id\" INTEGER REFERENCES \"guides_guide\"(\"id\")\n);CREATE TABLE \"blog_quotation\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"slug\" TEXT,\n   \"quotation\" TEXT,\n   \"source\" TEXT,\n   \"source_url\" TEXT,\n   \"created\" TEXT,\n   \"metadata\" TEXT,\n   \"import_ref\" TEXT,\n   \"card_image\" TEXT,\n   \"series_id\" TEXT REFERENCES \"blog_series\"(\"id\"),\n   \"is_draft\" INTEGER,\n   \"context\" TEXT\n);CREATE TABLE \"auth_user\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"password\" TEXT,\n   \"last_login\" TEXT,\n   \"is_superuser\" INTEGER,\n   \"username\" TEXT,\n   \"first_name\" TEXT,\n   \"last_name\" TEXT,\n   \"email\" TEXT,\n   \"is_staff\" INTEGER,\n   \"is_active\" INTEGER,\n   \"date_joined\" TEXT\n);CREATE TABLE \"blog_blogmark_tags\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"blogmark_id\" INTEGER REFERENCES \"blog_blogmark\"(\"id\"),\n   \"tag_id\" INTEGER REFERENCES \"blog_tag\"(\"id\")\n);CREATE TABLE \"blog_series\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"created\" TEXT,\n   \"slug\" TEXT,\n   \"title\" TEXT,\n   \"summary\" TEXT\n);CREATE TABLE \"django_content_type\" (\n   \"id\" INTEGER PRIMARY KEY,\n   \"app_label\" TEXT,\n   \"model\" TEXT\n);CREATE TABLE til(\n  path TEXT,\n  topic TEXT,\n  title TEXT,\n  url TEXT,\n  body TEXT,\n  html TEXT,\n  created TEXT,\n  created_utc TEXT,\n  updated TEXT,\n  updated_utc TEXT,\n  shot_hash TEXT,\n  slug TEXT,\n  summary TEXT\n)"}]
    ```

## Reasoning

**Counting entries**

I think the user wants to know how to count the blog entry rows, so I need to formulate a query that does just that. It seems straightforward: I’ll compose a request specifically aimed at counting those entries. This will help the user get the information they’re looking for without any unnecessary complexity. I really want to make sure I'm providing a clear and accurate response for them!

## Response

### Tool calls

- **Datasette_query**: `call_SYcRJnuCrQDRevxvd5TgK0Wd`  
    sqlite_sql:
    ```
    SELECT COUNT(*) AS count FROM blog_entry;
    ```

## Token usage

2,610 input, 49 output, `{"input_tokens_details": {"cache_write_tokens": 0, "cached_tokens": 0}, "output_tokens_details": {"reasoning_tokens": 18}}`

# 2026-07-30T00:13:56

## Prompt

-- none --

### Tools

- `Datasette({"url": "https://datasette.simonwillison.net/simonwillisonblog"})`:
    - **Datasette_query**: `1aa6d35`
    - **Datasette_schema**: `24ec615`

### Tool results

- **Datasette_query**: `call_SYcRJnuCrQDRevxvd5TgK0Wd`  
    ```
    [{"count": 3318}]
    ```

## Response

There are **3,318** entries in `blog_entry`.

## Token usage

2,680 input, 17 output, `{"input_tokens_details": {"cache_write_tokens": 0, "cached_tokens": 2048}, "output_tokens_details": {"reasoning_tokens": 0}}`

# 2026-07-30T00:14:04

## Prompt

now blogmarks​

### Tools

- `Datasette({"url": "https://datasette.simonwillison.net/simonwillisonblog"})`:
    - **Datasette_query**: `1aa6d35`
    - **Datasette_schema**: `24ec615`

## Response

### Tool calls

- **Datasette_query**: `call_n7I5vz25SwjE6nfoKUwzorlU`  
    sqlite_sql:
    ```
    SELECT COUNT(*) AS count FROM blog_blogmark;
    ```

## Token usage

2,669 input, 30 output, `{"input_tokens_details": {"cache_write_tokens": 0, "cached_tokens": 0}, "output_tokens_details": {"reasoning_tokens": 0}}`

# 2026-07-30T00:14:07

## Prompt

-- none --

### Tools

- `Datasette({"url": "https://datasette.simonwillison.net/simonwillisonblog"})`:
    - **Datasette_query**: `1aa6d35`
    - **Datasette_schema**: `24ec615`

### Tool results

- **Datasette_query**: `call_n7I5vz25SwjE6nfoKUwzorlU`  
    ```
    [{"count": 8422}]
    ```

## Response

There are **8,422** blogmarks in `blog_blogmark`.

## Token usage

2,720 input, 19 output, `{"input_tokens_details": {"cache_write_tokens": 0, "cached_tokens": 2560}, "output_tokens_details": {"reasoning_tokens": 0}}`

</details>

simonw:
I'm happy with this now. I'm going to merge to main.

Issue #1565, opened 2026-07-26
`llm openai endpoint` command for running (unlogged) prompts against arbitrary endpoints


simonw:
I often find myself wanting to exercise an OpenAI chat completions compatible endpoint from the terminal without having to configure it in `extra-openai-models.yaml` first.

A new command that lets you specify a URL directly and then run a prompt or conduct a chat would be useful.

It should not log to `logs.db` because we don't have a guaranteed unique model ID that we can use to tell the difference between the logs later on.

simonw:
Got this prototype working:
```bash
llm openai endpoint http://localhost:1234/v1 -m google/gemma-4-12b 'impress me with a joke'
```

simonw:
Plenty of notes on how I tested it in the PR:
- #1568

simonw:
Documentation for this new feature: https://llm.datasette.io/en/latest/other-models.html#openai-endpoint

PR #1568, opened 2026-07-27
llm openai endpoint command

simonw:
Refs:
- #1565


simonw:
This worked (against local LM Studio):

```bash
uv run --with llm-tools-quickjs --with-editable . \
  llm openai endpoint http://localhost:1234/v1 -m google/gemma-4-12b \
  -T QuickJS 'multiply 3434 * 2434' --td
```
Output:
```
Tool call: QuickJS_execute_javascript({'javascript': '3434 * 2434'})
  8358356


3434 multiplied by 2434 is 8,358,356.
```

simonw:
Oops, I opened this PR against the code in this other PR when it should have been against `main`:
- #1562

docs/help.md
```diff
@@ -1066,13 +1066,57 @@ Options:
...
+```
+Usage: llm openai endpoint [OPTIONS] URL [PROMPT]
+
+  Run against an OpenAI-compatible endpoint without logging.
+
+  If PROMPT is provided, execute it once. If PROMPT is omitted in an interactive
+  terminal, start a chat unless --template is provided. Templates run once by
+  default; use --chat to apply one interactively. Piped stdin is treated as a
+  one-off prompt. Use --models to list the available model IDs without running a
+  prompt.
```
simonw: I don't like that this behaves differently from regular `llm prompt` when you don't provide stdin or a prompt or template - I think it should do the same thing, hang awaiting input, and you should always have to pass --chat to get into a chat

llm/default_plugins/openai_models.py
```diff
@@ -401,9 +402,288 @@ def embed_batch(self, items: Iterable[str | bytes]) -> Iterator[list[float]]:
...
+        model_class = Responses if use_responses else Chat
+        model = model_class(
+            model_id=model_id or "",
+            model_name=model_id or "",
+            api_base=url,
+            headers=dict(headers),
+            vision=True,
+            audio=not use_responses,
+            supports_tools=True,
+        )
```
simonw: Here's the full list of class constructor parameters:

https://github.com/simonw/llm/blob/c2838a49d539a7a0dcb6309825ae23ace5a42ee5/llm/default_plugins/openai_models.py#L671-L691

simonw:
This doesn't support schemas yet. I think I'm going to hard-code `supports_schema=True` - if someone tries a prompt against an endpoint that doesn't support a schema they'll get an error, and that's OK.

llm/default_plugins/openai_models.py
simonw: I think we should default to True any of the options which won't break if the model turns out NOT to support that, but will throw an error if the user tries something that turns out not to be supported.

simonw: The tricky one here  is `reasoning=True`, which DOES change the JSON sent to the Responses API:

https://github.com/simonw/llm/blob/c2838a49d539a7a0dcb6309825ae23ace5a42ee5/llm/default_plugins/openai_models.py#L1313-L1320

simonw: The problem there is that `hide_reasoning` doesn't default to False, which means that if you pass `reasoning=True` to the class constructor for `Responses()` you will always end up with this added to the JSON:
```json
{"reasoning": {"summary": "auto"}}
```

simonw: I'm going to split that reasoning summary thing from reasoning, so our endpoint command can default to reasoning=True without default setting anything in the JSON unless the user passes a reasoning effort.

simonw:
With LM Studio running this:
```bash
uv run llm openai endpoint http://127.0.0.1:1234 --models
```
Returned blank. It should have returned an error.

This correctly returned a list of models:
```bash
uv run llm openai endpoint http://127.0.0.1:1234/v1 --models
```

simonw:
Testing again against LM Studio:
```bash
uv run llm openai endpoint http://127.0.0.1:1234/v1 \
  -m google/gemma-4-31b 'what is the current LLM version? And the time?' \
  -T llm_version -T llm_time --td
```
```
Tool call: llm_version({})
  0.32rc1


Tool call: llm_time({})
  {
    "utc_time": "2026-07-30 20:53:29 UTC",
    "utc_time_iso": "2026-07-30T20:53:29.194940+00:00",
    "local_timezone": "PDT",
    "local_time": "2026-07-30 13:53:29",
    "timezone_offset": "UTC-7:00",
    "is_dst": true
  }


The current LLM version is 0.32rc1, and the time is 1:53 PM PDT on July 30, 2026 (20:53 UTC).
```

docs/help.md
simonw: Fixed that in 859cea1e67c52df620d543ee84715e600f51956d

llm/default_plugins/openai_models.py
simonw: We now default to reasoning, verbosity, image_detail_original. All that's left is schemas.

simonw:
I tested `--chat` with tools and that worked great too.

```bash
uv run llm openai endpoint http://127.0.0.1:1234/v1 -m google/gemma-4-31b -T llm_version -T llm_time --td --chat 
```
```
Chatting with google/gemma-4-31b at http://127.0.0.1:1234/v1
...
```
And an attachment:
```bash
uv run llm openai endpoint http://127.0.0.1:1234/v1 -m google/gemma-4-31b -a ~/Downloads/waving.png 'describe image'
```
waving

Output:

> This image is a bright, colorful cartoon illustration featuring four cheerful white ducks riding small blue bicycles. The entire scene is set against a solid, vibrant magenta-pink background.
> 
> All four ducks are facing to the right and have happy expressions with wide, open beaks and brown eyes. Their bicycles are identical, with bright blue frames and black tires.
> 
> While they look very similar, there is a slight difference in their poses:
> * The **first (far left) and fourth (far right)** ducks have both wings gripping the handlebars of their bikes.
> * The **second and third** ducks have their right wing on the handlebar and their left wing raised high in the air as if waving.

simonw:
The `--models` lack of error bug is fixed:
```
% uv run llm openai endpoint http://127.0.0.1:1234/ --models
Error: Unexpected endpoint or method. (GET /models)
% uv run llm openai endpoint http://127.0.0.1:1234 --models 
Error: Unexpected endpoint or method. (GET /models)
% uv run llm openai endpoint http://127.0.0.1:1234/v1 --models
google/gemma-4-31b
ternary-bonsai-27b
bonsai-27b-mlx
ornith-1.0-35b
google/gemma-4-12b-qat
north-mini-code-1.0
google/gemma-4-12b
granite-4.1-30b
granite-4.1-8b-fp8
granite-4.1-3b
qwen3.6-35b-a3b
minimax-m2.7
google/gemma-4-26b-a4b
google/gemma-4-e4b
google/gemma-4-e2b
qwen3.5-27b-claude-4.6-opus-reasoning-distilled-v2
openai/gpt-oss-120b
qwen/qwen3-coder-next
qwen3.5-4b
qwen/qwen3.5-9b
qwen/qwen3.5-35b-a3b
text-embedding-nomic-embed-text-v1.5
```

simonw:
Schemas work too now:
```bash
uv run llm openai endpoint http://127.0.0.1:1234/v1 -m google/gemma-4-12b \
  --schema-multi 'name, age int, bio' 'invent 3 cool dogs'
```
```
{
  "items": [
    {
      "name": "The Cinder-Pelt",
      "age": 3,
      "bio": "A majestic, heavy-set dog with fur the color of cooled charcoal. Along its spine and tail, the hair glows with a soft orange ember light. These dogs are known for being incredibly loyal; they can sense their owner’s body temperature and will lay down beside them to provide warmth during cold nights."
    },
    {
      "name": "The Neon-Pulse",
      "age": 2,
      "bio": "A sleek, high-energy breed that looks like it stepped out of a cyberpunk movie. It has smooth, metallic-sheen fur and distinct 'circuit' patterns along its legs that pulse with neon blue light when the dog is excited or running."
    },
    {
      "name": "The Nebula-Stalker",
      "age": 5,
      "bio": "A mystical, ethereal breed whose fur appears to contain a swirling galaxy of deep purples and blues. It is said that these dogs can 'scent' memories rather than just smells, making them the ultimate companions for those seeking lost things or forgotten thoughts."
    }
  ]
}
```

simonw:
It's important that the endpoint command NEVER creates a logs.db database if the user does not have one yet. I want `uvx llm openai endpoint ...` to be a safe command to run that won't create a `.../io.datasette.llm` directory as a side-effect.

simonw:
Also manually tested against the new https://github.com/simonw/llm-chat-completion-server plugin:
```bash
llm chat-completions-server -p 9234
```
Then:
```bash
uv run llm openai endpoint http://127.0.0.1:9234/v1 -m gpt-5-nano --chat
```
That worked:
```
Chatting with gpt-5-nano at http://127.0.0.1:9234/v1
Type 'exit' or 'quit' to exit
Type '!edit' to open your default editor and modify the prompt
> tell me a joke about a pelican, 5 words only
Pelican opened bill, charges extra.
> another, 10 words now
Pelican at cafe charges for napkin, then swallows the bill.
```
And because it round-tripped through LLM (running the plugin) it got logged there:
```bash
llm logs -s
```
```yaml
- model: gpt-5-nano
  datetime: '2026-07-30T21:19:26'
  conversation: 01kytebbnbyw980kcv0x834epw
  prompt: tell me a joke about a pelican, 5 words onl
  prompt_fragments: []
  system_fragments: []
- model: gpt-5-nano
  datetime: '2026-07-30T21:19:42'
  conversation: 01kytebnbyzk5p1kqpkgd1f3qp
  prompt: another, 10 words now
  prompt_fragments: []
  system_fragments: []
```

PR #1571, opened 2026-07-28
Fix logs -f fragment filter returning no results on SQLite 3.51.0 and 3.51.1

### LINKED ISSUE #1511: llm logs -f fragment filter returns no results on SQLite 3.51.0+

ikatyal2110:
SQLite 3.51 changed how correlated subqueries inside EXISTS behave when the subquery body is a UNION: the outer correlation (`responses.id`) is silently dropped, causing the filter to return no rows even when matches exist. The fix rewrites `EXISTS (A UNION B)` as `(EXISTS (A) OR EXISTS (B))`, which is semantically identical but works correctly on all SQLite versions. Fixes #1511

---
_Generated by [Claude Code](https://claude.ai/code/session_01AFKMM42St23t57HSmS61Et)_

### DIFF
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -1978,19 +1978,22 @@ def logs_list(
 
         for i, fragment_hash in enumerate(fragment_hashes):
             exists_clause = f"""
-            exists (
-                select 1 from prompt_fragments
-                where prompt_fragments.response_id = responses.id
-                and prompt_fragments.fragment_id in (
-                    select fragments.id from fragments
-                    where hash = :f{i}
+            (
+                exists (
+                    select 1 from prompt_fragments
+                    where prompt_fragments.response_id = responses.id
+                    and prompt_fragments.fragment_id in (
+                        select fragments.id from fragments
+                        where hash = :f{i}
+                    )
                 )
-                union
-                select 1 from system_fragments
-                where system_fragments.response_id = responses.id
-                and system_fragments.fragment_id in (
-                    select fragments.id from fragments
-                    where hash = :f{i}
+                or exists (
+                    select 1 from system_fragments
+                    where system_fragments.response_id = responses.id
+                    and system_fragments.fragment_id in (
+                        select fragments.id from fragments
+                        where hash = :f{i}
+                    )
                 )
             )
             """


simonw:
Looks like this bug only affected SQLite 3.51.0 and 3.51.1 - they fixed it in 3.51.2. Here's the result of a test I had Codex run, comparing this PR to main:

SQLite | Original EXISTS(A UNION B) | PR rewrite
-- | -- | --
3.50.4 | r_prompt | r_prompt
3.51.0 | no rows | r_prompt
3.51.1 | no rows | r_prompt
3.51.2 | r_prompt | r_prompt
3.53.4 | r_prompt | r_prompt

https://sqlite.org/releaselog/3_51_2.html says:

> Fix multiple problems in the EXISTS-to-JOIN optimization that was added as part of optimization item 6b above.

simonw:
Landed a version of this that resolved the conflicts here: 9efdfa6d5836011e4d9c7071a500e0e503368164

PR #1575, opened 2026-07-30
Fix pending_tool_calls to exclude server-executed tool calls

### LINKED ISSUE #1574: LogStore.pending_tool_calls returns server-executed tool calls as pending
LogStore.pending_tool_calls in llm/logs.py returns every ToolCallPart at the chain tip, including those with server_executed=True. Server-executed calls have already been handled by the provider and should not be treated as work waiting for the client. The parallel function _trailing_pending_tool_calls in models.py already filters these out with `not p.server_executed`, but pending_tool_calls in the new LogStore does not.

ikatyal2110:
LogStore.pending_tool_calls was returning ToolCallParts with server_executed=True as pending work for the client, even though the provider has already handled them. The fix adds the same `not part.server_executed` guard that _trailing_pending_tool_calls in models.py already applies, along with a regression test covering this case. Fixes #1574

---
_Generated by [Claude Code](https://claude.ai/code/session_01522mwpRNpQJ9XSwTRhCr7F)_

### DIFF
--- a/llm/logs.py
+++ b/llm/logs.py
@@ -641,7 +641,11 @@ class LogStore:
         chain = self.load_chain(tip)
         if not chain:
             return []
-        return [part for part in chain[-1].parts if isinstance(part, ToolCallPart)]
+        return [
+            part
+            for part in chain[-1].parts
+            if isinstance(part, ToolCallPart) and not part.server_executed
+        ]
 
 
 def ensure_attachment(db, attachment) -> str:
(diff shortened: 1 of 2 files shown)


simonw:
Thanks, this is a good catch.

simonw:
Demo of the bug. With the current `main`:
```bash
uv run python -c '
import llm, sqlite_utils
from llm.logs import LogStore
from llm.parts import ToolCallPart
s=LogStore(sqlite_utils.Database(memory=True))
tip=s.ensure_chain([llm.user("Search"), llm.assistant(ToolCallPart(name="web_search", arguments={"q":"pelicans"}, tool_call_id="tc1", server_executed=True))])
print([p.tool_call_id for p in s.pending_tool_calls(tip)])'
```
Outputs:
```
['tc1']
```
After this fix, it outputs:
```
[]
```

PR #1578, opened 2026-07-31
Narrow except ValueError in embed-multi to only handle missing-model case

### LINKED ISSUE #1523: Embeddings migrations fail with sqlite-utils 4.0; embed-multi masks it as "You need to specify an embedding model"

ikatyal2110:
The broad `except ValueError` around `Collection()` in the `embed-multi` command would silently convert any ValueError (e.g. from a database migration failure) into the misleading "You need to specify an embedding model" message. The fix extracts the model ID first, then only shows the user-friendly message when no model was provided; all other ValueErrors are re-raised. A regression test verifies that a non-model ValueError propagates instead of being masked. Fixes #1523

---
_Generated by [Claude Code](https://claude.ai/code/session_01P21wrXuJbMCwW8gNsYoWVM)_

### DIFF
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -3363,7 +3363,7 @@ def embed_multi(
     effective_model_id = model or get_default_embedding_model()
     try:
         collection_obj = Collection(collection, db=db, model_id=effective_model_id)
-    except ValueError as e:
+    except ValueError:
         if effective_model_id is None:
             raise click.ClickException(
                 "You need to specify an embedding model (no default model is set)"
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -3360,14 +3360,15 @@ def embed_multi(
     for alias, attach_path in attach:
         db.attach(alias, attach_path)
 
+    effective_model_id = model or get_default_embedding_model()
     try:
-        collection_obj = Collection(
-            collection, db=db, model_id=model or get_default_embedding_model()
-        )
-    except ValueError:
-        raise click.ClickException(
-            "You need to specify an embedding model (no default model is set)"
-        )
+        collection_obj = Collection(collection, db=db, model_id=effective_model_id)
+    except ValueError as e:
+        if effective_model_id is None:
+            raise click.ClickException(
+                "You need to specify an embedding model (no default model is set)"
+            )
+        raise
 
     expected_length = None
     if files:
--- a/tests/test_embed_cli.py
+++ b/tests/test_embed_cli.py
@@ -1,7 +1,7 @@
 import json
 import pathlib
 import sys
-from unittest.mock import ANY
+from unittest.mock import ANY, patch
 
 import pytest
 import sqlite_utils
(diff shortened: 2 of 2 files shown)


llm/cli.py
```diff
@@ -3360,14 +3360,15 @@ def embed_multi(
...
     try:
-        collection_obj = Collection(
-            collection, db=db, model_id=model or get_default_embedding_model()
-        )
-    except ValueError:
-        raise click.ClickException(
-            "You need to specify an embedding model (no default model is set)"
-        )
+        collection_obj = Collection(collection, db=db, model_id=effective_model_id)
+    except ValueError as e:
```
Sanjays2402: the same masking exists in `embed` a couple hundred lines up: if `-m` names a model that isn't installed, `get_embedding_model` raises UnknownModelError and the handler turns it into "You need to specify an embedding model (no default model is set)", which is confusing when you did specify one. worth the same `if model is None` guard there so an unknown-model error surfaces as itself.

simonw:
I swapped ValueError for UnknownModelError instead. https://github.com/simonw/llm/commit/f3f4f29c6e5e623c3776cbf6455dda2345638214

PR #1579, opened 2026-07-31
Ollama OCR with URLs fails where filepaths succeed

ojassharma7:
Fixes:
- #1046.

## What changed
- `llm/models.py`
- `tests/test_attachments.py`

## Verification
The project's own test suite was run before and after this change; it introduces no new test failures or lint violations.


simonw:
Thanks for this. Looks like the default `https` max redirects is 20, I'm going to drop that to 3 for this feature.

PR #1581, opened 2026-08-01
Skip empty toolboxes in text output of `llm tools`

### LINKED ISSUE #1580: `llm tools` looks broken when listing dynamic toolboxes
With https://github.com/simonw/llm-mcp-client installed:
```bash
llm tools
```
Outputs:
```
llm_time() -> dict (plugin: llm.default_plugins.default_tools)

  Returns the current time, as local time and UTC

llm_version() -> str (plugin: llm.default_plugins.default_tools)

  Return the installed version of llm

MCP:
```
This is technically correct, since `MCP` only returns tools if you call it with a specific URL e.g. `MCP("https://datasette.simonwillison.net/-/mcp")` - but it doesn't look good.

ikatyal2110:
When a plugin registers a toolbox class whose `method_tools()` yields no tools — for example an MCP client that only produces tools once a URL is supplied — `llm tools` printed the toolbox name (e.g. `MCP:`) followed by nothing, which looked broken. This change skips such empty toolboxes in the human-readable text output; JSON output still lists them with `"tools": []` for programmatic consumers. Fixes #1580

---
_Generated by [Claude Code](https://claude.ai/code/session_01CfUpdgpMBXSgfj9kUg3KES)_

### DIFF
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -2764,8 +2764,11 @@ def tools_list(tool_defs, json_, python_tools):
             if tool.description:
                 click.echo(textwrap.indent(tool.description.strip(), "  ") + "\n")
         for toolbox in toolbox_objects:
+            toolbox_method_tools = list(toolbox.method_tools())
+            if not toolbox_method_tools:
+                continue
             click.echo(toolbox.name + ":\n")
-            for tool in toolbox.method_tools():
+            for tool in toolbox_method_tools:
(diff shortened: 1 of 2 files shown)


simonw:
I fixed this another way in https://github.com/simonw/llm/commit/1b99533ff83aaed70a74dd70781dd4329fb350bb

PR #1584, opened 2026-08-02
Fix template schema_object overriding --schema command-line argument

### LINKED ISSUE #1583: Template schema_object overwrites --schema command-line argument instead of being a default
When a template defines `schema_object` and the user also passes `--schema` on the command line, the template's schema silently wins instead of the command-line value.

This is inconsistent with every other template setting: `model` is only applied `if model_id is None`, `extract`/`extract_last` are only applied `if not (extract or extract_last)`, and options are merged with `_merge_template_options` which skips template values that were already specified. The `schema_object` check at line 903-904 of `cli.py` is the only one that unconditionally overwrites whatever the command line provided.

ikatyal2110:
When a template defines `schema_object` and the user also passes `--schema` on the command line, the template's schema was silently winning instead of the command-line value. Every other template setting (model, extract, options) only applies when the command line did not supply a value, so this one-line fix makes `schema_object` consistent with that pattern. Fixes #1583.

---
_Generated by [Claude Code](https://claude.ai/code/session_01T2B2Q3r6MjX2PQXHQAj4ox)_

### DIFF
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -900,7 +900,7 @@ def prompt(
             fragments = [*template_obj.fragments, *fragments]
         if template_obj.system_fragments:
             system_fragments = [*template_obj.system_fragments, *system_fragments]
-        if template_obj.schema_object:
+        if template_obj.schema_object and not schema:
             schema = template_obj.schema_object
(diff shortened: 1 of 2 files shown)


simonw:
This looks to be a duplicate of:
- #1588

PR #1585, opened 2026-08-02
Add service_tier option for OpenAI models

simonw:
OpenAI [renamed Priority processing to Fast mode](https://openai.com/api-fast-mode/) on July 30th 2026, and `gpt-5.6-sol` supports it for up to 2.5x faster processing at a higher per-token price. It is controlled by the [`service_tier`](https://developers.openai.com/api/docs/guides/fast-mode) API parameter, accepted by both the Responses and Chat Completions endpoints.

This adds a `service_tier` option to OpenAI models registered with a new `service_tier=True`:

```bash
llm -m gpt-5.6-sol -o service_tier fast 'Fast facts about pelicans'
```
This is available for other models too, especially since `service_tier: flex` looks to be supported by many of them (half price in exchange for slower and less reliable.)

<details><summary>Claude Code PR (Fable 5)</summary>

## Design notes

- The option is a **plain string, not an enum** - the value is passed straight through to the API, so `fast`, `priority`, `flex` and any future tier names work on other models/endpoints without code changes.
- The flag is enabled for `gpt-5.6-sol` only (the one model OpenAI documents as supporting Fast mode). Custom models can opt in with `service_tier: true` in `extra-openai-models.yaml`.
- The option flows through both code paths: the Responses API and the `-o chat_completions 1` fallback (the delegate constructor now carries the flag across).
- The requested tier is persisted in the logged options (`turns.options_json`), so `llm logs -c --json` shows it. The tier that *actually* processed the request (echoed by the API, since Fast mode can fall back to standard processing) is available via `response.json()["service_tier"]` in the Python API but is not persisted by the new logging schema - the docs note this.

## Documentation

New "Fast mode and service tiers" section in `docs/openai-models.md` linking to OpenAI's official Fast mode documentation, plus a note in the extra-models section and cog-regenerated options listing in `docs/usage.md`.

## Testing

- 7 new tests: option exposed on `gpt-5.6-sol` but not Terra/Luna/5.5, kwargs building, and request-body assertions for both the `/v1/responses` and `/v1/chat/completions` paths via httpx mocks (including the echoed tier surfacing in `response.json()`).
- Full suite passes: 979 tests. ruff, black and `cog --check` clean.
- Manually tested against the live API: `-o service_tier fast` was accepted, the response echoed `service_tier: priority` (OpenAI's documented echo for GPT-5.6-era models), and timing the same 12-stanza-poem prompt twice per tier showed ~1.37x output tokens/sec over standard.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01CYbKrezWv4ANYuXGFjJGkN

</details>

https://claude.ai/code/session_01CYbKrezWv4ANYuXGFjJGkN

### DIFF
--- a/llm/default_plugins/openai_models.py
+++ b/llm/default_plugins/openai_models.py
@@ -38,39 +38,82 @@ from llm.utils import (
 def register_models(register):
     # GPT-4o
     register(
-        Chat("gpt-4o", vision=True, supports_schema=True, supports_tools=True),
-        AsyncChat("gpt-4o", vision=True, supports_schema=True, supports_tools=True),
+        Chat(
+            "gpt-4o",
+            vision=True,
+            service_tier=True,
+            supports_schema=True,
+            supports_tools=True,
+        ),
+        AsyncChat(
+            "gpt-4o",
+            vision=True,
+            service_tier=True,
+            supports_schema=True,
+            supports_tools=True,
+        ),
         aliases=("4o",),
     )
     register(
-        Chat("gpt-4o-mini", vision=True, supports_schema=True, supports_tools=True),
+        Chat(
+            "gpt-4o-mini",
+            vision=True,
+            service_tier=True,
+            supports_schema=True,
+            supports_tools=True,
+        ),
         AsyncChat(
-            "gpt-4o-mini", vision=True, supports_schema=True, supports_tools=True
+            "gpt-4o-mini",
+            vision=True,
+            service_tier=True,
+            supports_schema=True,
+            supports_tools=True,
         ),
         aliases=("4o-mini",),
     )
     # GPT-4.1
     for model_id in ("gpt-4.1", "gpt-4.1-mini", "gpt-4.1-nano"):
         register(
-            Chat(model_id, vision=True, supports_schema=True, supports_tools=True),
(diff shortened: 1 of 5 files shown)


llm/default_plugins/openai_models.py
```diff
@@ -250,13 +250,16 @@ def register_models(register):
 
     # GPT-5.6
     for model_id in ("gpt-5.6-sol", "gpt-5.6-terra", "gpt-5.6-luna"):
+        # gpt-5.6-sol supports service_tier for Fast mode
+        service_tier = model_id == "gpt-5.6-sol"
```
simonw: It's not clear from the OpenAI docs if GPT-5.6 is the only model to support fast mode. Also Flex mode exists and I think that works on all models.

I'm going to enable this for every OpenAI model instead.

simonw:
I finally found the docs that tell you which models support fast and flex mode: it's on the pricing page behind this toggle:

CleanShot 2026-08-03 at 12 34 24@2x

https://developers.openai.com/api/docs/pricing?latest-pricing=standard

This supports my decision to make `service_tier` available on all OPenAI models (except old `/v1/completions` GPT 3.5 turbo instruct.)

PR #1588, opened 2026-08-03
Fix llm openai endpoint ignoring --schema when template also defines schema_object

### LINKED ISSUE #1583: Template schema_object overwrites --schema command-line argument instead of being a default

ikatyal2110:
The `llm openai endpoint` command applied `template.schema_object` unconditionally at `openai_models.py:587`, overwriting any `--schema` value the user explicitly passed on the command line. This is inconsistent with how every other template override works: `model` is only applied `if not model_id`, options use `_merge_template_options` (skipping already-specified keys), and attachments are additive. The same bug was fixed for the main `prompt` command in #1584, but the `endpoint` command has its own copy of the template-application block that was not included in that fix.

Fix: add `and not schema` to the guard condition, matching the pattern used everywhere else. The existing `test_endpoint_template` already invoked the command with both `--template` (schema_object set) and `--schema`; its assertion was corrected to expect the CLI value to win. A companion test confirms the template schema_object is still applied when no `--schema` is given.

Fixes #1583

### DIFF
--- a/llm/default_plugins/openai_models.py
+++ b/llm/default_plugins/openai_models.py
@@ -584,7 +584,7 @@ def register_commands(cli):
                 raise click.ClickException(str(ex))
             if not model_id and template_obj.model:
                 model_id = template_obj.model
-            if template_obj.schema_object:
+            if template_obj.schema_object and not schema:
                 schema = template_obj.schema_object
             if template_obj.options:
                 options = _merge_template_options(template_obj, options)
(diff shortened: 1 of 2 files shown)


simonw:
Thanks for this.

PR #1596, opened 2026-08-06
Fix Anthropic API key alias in Python API docs

ellaclairebuild:
## Summary
- Update the llm-anthropic Python API example to reference `llm keys set anthropic` instead of the legacy `claude` key alias.
- Aligns with `docs/index.md` and the llm-anthropic plugin (`needs_key = "anthropic"`).

## Test plan
- [x] Docs-only change in `docs/python-api.md`

### DIFF
--- a/docs/python-api.md
+++ b/docs/python-api.md
@@ -543,7 +543,7 @@ Then in your Python code:
 import llm
 
 model = llm.get_model("claude-3.5-sonnet")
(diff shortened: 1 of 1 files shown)


simonw:
This PR rewrites the entire file (looks like `\r` line endings). I'll apply the `llm keys set anthropic ` line independently of this.

PR #1603, opened 2026-08-09
Add !attach support to interactive chat

simonw:
`!attach url-or-path` in chat.

<details><summary>Codex PR</summary>
## Summary

- add `!attach <url-or-path>` to interactive chat sessions
- support attachments alongside `!multi` and `!fragment`
- use the shared chat loop, making the command available to `llm chat` and `llm openai endpoint --chat`
- document the command and test attachment delivery and logging

## Why

Interactive chat sessions had no command for attaching a file or URL to a later turn. This implements the same general feature proposed in #734, updated for the current shared chat loop and message logging architecture.

## Validation

- `uv run pytest -q tests/test_chat.py tests/test_openai_endpoint.py` — 37 passed
- `uv run black --check llm/cli.py tests/test_chat.py`
- `uv run ruff check llm/cli.py tests/test_chat.py`
</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,7 @@ Type 'exit' or 'quit' to exit
 Type '!multi' to enter multiple lines, then '!end' to finish
 Type '!edit' to open your default editor and modify the prompt.
 Type '!fragment <my_fragment> [<another_fragment> ...]' to insert one or more fragments
+Type '!attach <url-or-path>' to attach a file
 > Tell me a joke about a pelican
 Why don't pelicans like to tip waiters?
 
--- a/docs/fragments.md
+++ b/docs/fragments.md
@@ -58,6 +58,7 @@ Type 'exit' or 'quit' to exit
 Type '!multi' to enter multiple lines, then '!end' to finish
 Type '!edit' to open your default editor and modify the prompt.
 Type '!fragment <my_fragment> [<another_fragment> ...]' to insert one or more fragments
+Type '!attach <url-or-path>' to attach a file
 > Explain this document to me
 ```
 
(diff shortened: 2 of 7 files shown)


simonw:
Doesn't feel quite right. It sends the attachment with an empty prompt as soon as you hit enter, so you don't get a chance to say what you want to do with it:
```
% uv run llm chat
Chatting with gpt-5.6-luna
Type 'exit' or 'quit' to exit
Type '!edit' to open your default editor and modify the prompt
Type '!fragment <my_fragment> [<another_fragment> ...]' to insert one or more fragments
Type '!attach <url-or-path>' to attach a file
> !attach https://static.simonwillison.net/static/2026/hy3-pelican.png
 
A cartoon stork riding a red bicycle against a light blue background. The bicycle has black wheels, orange pedals, and the stork’s long neck extends upward to its white head and large orange beak.
```
Related issue: if the URL is invalid it quits the whole chat session (idea here was what if you put a prompt after the attach, but that doesn't make sense because a filename could have spaces in it):
```
> !attach https://static.simonwillison.net/static/2026/hy3-pelican.png write a haiku about it
Error: Client error '404 Not Found' for url 'https://static.simonwillison.net/static/2026/hy3-pelican.png%20write%20a%20haiku%20about%20it'
For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404
```

simonw:
Fixed in 8040274.

- A standalone `!attach` now queues the attachment and returns to the prompt instead of sending an empty model request.
- Multiple standalone attachments accumulate and are consumed by the next real message only.
- Invalid paths or URLs print an error without terminating the chat or discarding attachments already queued.
- Pending command-line fragments and attachments now also survive entering `!multi`.

The regression test covers queueing multiple files, continuing after an invalid attachment, attaching them to the next prompt, and not leaking them into the following turn.

PR #1604, opened 2026-08-10
fix: guard empty aliases list in 'llm aliases list'

### LINKED ISSUE #1602: `llm aliases list` crashes with ValueError when no aliases are defined
Running `llm aliases list` (without `--json`) on a fresh install or after removing all aliases crashes with `ValueError: max() arg is an empty sequence` because the code calls `max()` on an empty list before checking whether there is anything to display.

The `--json` variant handles the empty case correctly (returns `{}`), but the plain text path does not guard against it.

abhi-0203:
## Summary

Fixes a `ValueError: max() arg is an empty sequence` when running `llm aliases list` on a fresh install or after removing all aliases.

The `--json` variant already handles the empty case correctly (returns `{}`), but the plain text path calls `max()` on an empty sequence before checking whether there is anything to display.

## Fix

Add an early return when no aliases exist, before the `max()` call. This matches the `--json` path's behavior.

## Reproduction

```bash
llm aliases list  # on fresh install with no aliases defined
# ValueError: max() arg is an empty sequence
```

Closes #1602

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -4,6 +4,7 @@
 ## Unreleased
 
 - Reasoning-capable Responses API models now support a `reasoning_summary` option with `auto`, `concise`, and `detailed` values. This can be used with {ref}`llm openai endpoint --responses <openai-endpoint>`. [#1600](https://github.com/simonw/llm/issues/1600)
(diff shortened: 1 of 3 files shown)


simonw:
Adding a test, then I'll land this.

PR #1605, opened 2026-08-10
feat: add compare command for comparing model responses

SrikarManthatti:
## Summary

Adds a new `llm compare` command that runs the same prompt against two or more models and displays their responses for easy comparison.

## What's included

* Add `llm compare` with support for multiple `-m/--model` arguments.
* Run model requests concurrently while preserving the order of the supplied models.
* Display two model responses side-by-side and three or more responses sequentially.
* Add `--json` output for machine-readable comparison results.
* Support system prompts, model options, attachments, fragments, and prompts provided through stdin.
* Handle model-specific errors without preventing other comparison results from being displayed.
* Log successful comparison responses using the existing `llm` logging functionality.
* Add documentation and tests covering command usage, output formats, input handling, error cases, and logging.

This provides a simple way to evaluate and compare multiple models using the same input directly from the command line.

### DIFF
--- a/README.md
+++ b/README.md
@@ -397,6 +397,7 @@ For everything else, see [the llm tag](https://simonwillison.net/tags/llm/) on m
     * [llm embed-models –help](https://llm.datasette.io/en/stable/help.html#llm-embed-models-help)
     * [llm collections –help](https://llm.datasette.io/en/stable/help.html#llm-collections-help)
     * [llm openai –help](https://llm.datasette.io/en/stable/help.html#llm-openai-help)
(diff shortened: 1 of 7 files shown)


SrikarManthatti:
Hi @simonw - just wanted to bring this PR to your attention when you get a chance. It adds a new `compare` feature for comparing multiple LLM models side by side. 
Would really appreciate your feedback whenever you have some time. Thanks!

simonw:
This would make a good plugin but I don't think it should be part of LLM core.

It can use this plugin hook: https://llm.datasette.io/en/stable/plugins/plugin-hooks.html#register-commands-cli

Issue #1608, opened 2026-08-12
Fresh install broken and test suite escapes to live API since openai 3.0.0 (httpx → httpx2)


aborruso:
`openai` 3.0.0 was released today (2026-08-12) and its only breaking change is [the migration to HTTPX2](https://github.com/openai/openai-python/blob/main/httpx2.md): it now depends on `httpx2<3,>=2.7.0` where 2.x depended on `httpx<1,>=0.23.0`. `llm` requires `openai>=2.32.0` with no upper bound, so a clean resolve today gets `openai==3.0.0` and no `httpx` at all.

That has two separate consequences.

## 1. Every fresh install is broken

`httpx` is imported in `llm/models.py`, `llm/cli.py`, `llm/utils.py` and `llm/default_plugins/openai_models.py`, but it is not declared in `dependencies` — it used to arrive transitively via `openai`. So any command now dies at import time:

```
$ uv tool install llm
$ llm models
Traceback (most recent call last):
  File ".../bin/llm", line 4, in <module>
    from llm.cli import cli
  File ".../site-packages/llm/__init__.py", line 11, in <module>
    from .embeddings import Collection
  File ".../site-packages/llm/embeddings.py", line 13, in <module>
    from .models import EmbeddingModel
  File ".../site-packages/llm/models.py", line 31, in <module>
    import httpx
ModuleNotFoundError: No module named 'httpx'
```

Reproduction on current `main` (c410af0), installing without the dev group:

```
uv venv v1
VIRTUAL_ENV=v1 uv pip install .
v1/bin/llm models     # ModuleNotFoundError: No module named 'httpx'
```

The fix is one line in `pyproject.toml`:

```diff
 dependencies = [
     "click",
     "condense-json>=1.1",
+    "httpx",
     "openai>=2.32.0",
```

With that, the same clean install resolves `openai==3.0.0`, `httpx2==2.10.0` and `httpx==0.28.1` side by side, and `llm models` works again.

This does not appear to require any httpx2 migration in `llm` itself. The only place an httpx object reaches the OpenAI client is `openai_models.py:1336`, where `LLM_OPENAI_SHOW_RESPONSES` passes `http_client=logging_client()`, and openai 3.0.0's legacy escape hatch still accepts a plain `httpx.Client`. I checked that the custom transport is genuinely used rather than silently ignored:

```python
# uv run --no-project --with 'openai==3.0.0' --with httpx python probe.py
import httpx, openai

PAYLOAD = {
[...]

PR #1609, opened 2026-08-12
Declare httpx as a dependency

aborruso:
Refs #1608.

`httpx` is imported in `llm/models.py`, `llm/cli.py`, `llm/utils.py` and `llm/default_plugins/openai_models.py`, but it was never declared in `dependencies` — it arrived transitively through `openai`. `openai` 3.0.0 (released 2026-08-12) migrated to HTTPX2, so a fresh install no longer gets `httpx` and every command dies at import time with `ModuleNotFoundError: No module named 'httpx'`.

Verified both directions with a clean install, without the dev group (which would mask the problem, since `pytest-httpx` brings `httpx` in):

```
uv venv v1 && VIRTUAL_ENV=v1 uv pip install .
v1/bin/llm models
```

| tree | result |
| --- | --- |
| `main` (c410af0) | `ModuleNotFoundError: No module named 'httpx'` |
| this branch | works — resolves `openai==3.0.0`, `httpx2==2.10.0`, `httpx==0.28.1` |

No httpx2 migration seems necessary in `llm` itself: the only place an httpx object reaches the OpenAI client is `openai_models.py:1336` (`LLM_OPENAI_SHOW_RESPONSES` → `http_client=logging_client()`), and openai 3.0.0's legacy escape hatch still accepts a plain `httpx.Client` — #1608 has a runnable check confirming the custom transport is really used, not silently ignored.

This only covers the broken install. The second half of #1608 — `pytest-httpx` no longer intercepting openai's traffic under httpx2, so tests reach `api.openai.com` for real — is untouched here, since pinning versus migrating to `pytest-httpx2` is your call.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,6 +27,7 @@ classifiers = [
 dependencies = [
     "click",
     "condense-json>=1.1",
+    "httpx",
     "openai>=2.32.0",
     "click-default-group>=1.2.3",
     "sqlite-utils>=4.0",


simonw:
Handled by:
- #1631
- #1630

aborruso:
Thank you @simonw

PR #1612, opened 2026-08-13
fix: accept JSON Schema type names in schema_dsl

### LINKED ISSUE #1607: schema_dsl() silently maps full JSON Schema type names (integer, number, boolean, string) to string
Passing a full JSON Schema type name to `schema_dsl()` silently produces `"type": "string"` for every field instead of the intended type.

**Repro:**
```python
from llm.utils import schema_dsl
schema_dsl('count integer')
# {'type': 'object', 'properties': {'count': {'type': 'string'}}, 'required': ['count']}
# Expected: {'type': 'object', 'properties': {'count': {'type': 'integer'}}, 'required': ['count']}

schema_dsl('score float')        # → type: string (should be number)
schema_dsl('active boolean')     # → type: string (should be boolean)
schema_dsl('name string')        # → type: string (correct by accident)
```

**Root cause** (`llm/utils.py` lines 366–371): `type_mapping` only contains Python-style abbreviations (`int`, `float`, `bool`, `str`). Full JSON Schema names — `integer`, `number`, `boolean`, `string` — are not in the mapping, so the default `"string"` is used silently instead of raising or mapping to the correct type.

The fix is to add the four full names as pass-through entries in `type_mapping`.

Kayforkind:
## Summary

- `schema_dsl()` only mapped Python-style abbreviations (`int`, `float`, `bool`, `str`). Full JSON Schema names (`integer`, `number`, `boolean`, `string`) fell through to the default `"string"`.
- Map those four names through to the matching JSON Schema types. Short forms are unchanged.
- Document the aliases next to the existing type list.

Fixes #1607

## What I changed

- Files: `llm/utils.py`, `tests/test_utils.py`, `docs/schemas.md`
- Behavior before → after: `schema_dsl('count integer')` produced `{"type": "string"}`; it now produces `{"type": "integer"}`. Same for `number` / `boolean` / `string`.

## Test plan

- [x] Reproduced the issue: `schema_dsl('count integer')` returned string before the mapping change
- [x] Project test command: `uv run pytest tests/test_utils.py` (82 passed)
- [x] Manual check: `schema_dsl('count integer')`, `'score number'`, `'active boolean'`, `'name string'`, and `'age int'` all return the expected types

## Notes for maintainers

- I can iterate on review comments

### DIFF
--- a/docs/schemas.md
+++ b/docs/schemas.md
@@ -472,7 +472,7 @@ You can include type information by adding a type indicator after the property n
 
     name, bio, age int
 
-Supported types are `int` for integers, `float` for floating point numbers, `str` for strings (the default) and `bool` for true/false booleans.
+Supported types are `int` or `integer` for integers, `float` or `number` for floating point numbers, `str` or `string` for strings (the default) and `bool` or `boolean` for true/false booleans.
 
 To include a description of the field to act as a hint to the model, add one after a colon:
 
--- a/llm/utils.py
+++ b/llm/utils.py
@@ -365,9 +365,13 @@ def schema_dsl(schema_dsl: str, multi: bool = False) -> dict[str, Any]:
     # Type mapping dictionary
     type_mapping = {
         "int": "integer",
+        "integer": "integer",
         "float": "number",
+        "number": "number",
         "bool": "boolean",
+        "boolean": "boolean",
         "str": "string",
+        "string": "string",
     }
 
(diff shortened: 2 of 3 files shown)


simonw:
I went a different direction in https://github.com/simonw/llm/commit/307bd0e540b393f16d33fbfe3566b2f12ae154b1

PR #1619, opened 2026-08-17
Deduplicate required in schema_dsl()

### LINKED ISSUE #1616: schema_dsl() emits a duplicate in required and silently drops a field when a property name is repeated
Repeating a property name in the concise schema syntax emits a JSON schema that is invalid against the JSON Schema meta-schema, and silently drops the earlier field.

`schema_dsl()` writes properties into a dict but appends to `required`, so a repeated name collapses in one place and accumulates in the other:

```python
json_schema["properties"][field_name] = {"type": field_type}   # overwrites
...
json_schema["required"].append(field_name)                     # accumulates
```

## Reproduction

Using the command the docs recommend for experimenting with the syntax:

```bash
llm schemas dsl 'name, age int, name'
```

```json
{
  "type": "object",
  "properties": {
    "name": {"type": "string"},
    "age": {"type": "integer"}
  },
  "required": ["name", "age", "name"]
}
```

`required` has `name` twice. The 2020-12 validation meta-schema defines it as:

```json
"required": { "$ref": "#/$defs/stringArray" },

"stringArray": {
    "type": "array",
    "items": { "type": "string" },
    "uniqueItems": true,
    "default": []
}
```

so `uniqueItems` makes that array invalid. `jsonschema` agrees:

```
Draft202012Validator.check_schema(...)
-> ['name', 'age', 'name'] has non-unique elements
```

## The shape that is easier to hit

A long newline-separated schema where a name gets reused. Here the first field is lost as well as the schema being invalid:

```bash
llm schemas dsl 'name: first name
name: last name'
```

```json
{
  "type": "object",
  "properties": {
    "name": {"type": "string", "description": "last name"}
  },
  "required": ["name", "name"]
}
```

Two properties were asked for, one came back, and nothing said so. `--schema-multi` carries the same `required` array down inside `items`.

## Where it ends up

I could not find any schema validation in `llm/`:

```
grep -rn "check_schema\|validate(\|Draft20\|jsonschema" llm/*.py
(no matches)
```

so the schema goes to the provider as-is. I have not tested what a live provider does with it, and I would rather not guess; the point is that the local step that could catch it does not, and the failure surfaces somewhere further away than the typo that caused it.

## Two ways to fix, and I do not want to presume

1. **Reject it.** Raise a `ValueError` naming the repeated field, in the spirit of the malformed-field error added in `8dfd6b4`. A repeated name is almost always a mistake, and the DSL has no syntax for redefining a field.
2. **Make it consistent.** Keep last-wins for properties and deduplicate `required` so the output is at least a valid schema. Quieter, but it keeps the silent field loss.

I lean towards the first, since the second still throws away a field the user wrote. Happy to send a PR for whichever you prefer.

Related, but different causes and already filed, so noting only so this is not read as a duplicate: #1607 (unrecognised type names silently becoming strings) and #1545 (missing name before `:`).

## Environment

`llm` 0.32, `main` at `5b332ec`, Python 3.12.12. Nothing here is platform specific.

arose26:
Closes #1616

`schema_dsl()` writes each field into `properties` (a dict, so a repeated name overwrites) but appends every field to `required` (a list, so a repeated name accumulates):

```python
json_schema["properties"][field_name] = {"type": field_type}
...
json_schema["required"].append(field_name)
```

So a repeated property name produces a schema that is invalid against the JSON Schema meta-schema:

```
llm schemas dsl 'name, age int, name'
```

```json
{
  "type": "object",
  "properties": {"name": {"type": "string"}, "age": {"type": "integer"}},
  "required": ["name", "age", "name"]
}
```

`required` is a `stringArray` in 2020-12, which carries `uniqueItems: true`:

```python
>>> from jsonschema import Draft202012Validator
>>> Draft202012Validator.check_schema(schema_dsl("name, age int, name"))
jsonschema.exceptions.SchemaError: ['name', 'age', 'name'] has non-unique elements
```

The newline form is the easier one to hit by accident, since a long schema can reuse a name without it being obvious:

```
name: first name
name: last name
```

which yields one property and `"required": ["name", "name"]`.

### Change

A name already in `required` is not appended again. Both examples above now pass `check_schema()`.

### On the dropped field

The issue also notes that the earlier definition is silently discarded when a name repeats. That is left as is — last-one-wins is what assigning into `properties` already does, and rejecting duplicate names outright would be a behaviour change for anyone whose schema currently relies on it. Test case 10 pins the current behaviour explicitly so it is a decision rather than an accident; happy to make it raise instead if you would prefer that.

### Tests

Two cases added to the existing `test_schema_dsl` parametrize list:

- a repeated name in the comma form is listed once in `required`
- a repeated name in the newline form keeps the last description and is listed once

Both fail before this change and pass after. `tests/test_utils.py` passes (83 tests), and `ruff check` and `black --check` are clean on both files.

### DIFF
--- a/llm/utils.py
+++ b/llm/utils.py
@@ -413,8 +413,9 @@ def schema_dsl(schema_dsl: str, multi: bool = False) -> dict[str, Any]:
         if description:
             json_schema["properties"][field_name]["description"] = description
 
-        # Add field to required list
-        json_schema["required"].append(field_name)
+        # Add field to required list, which repeating a name must not duplicate
+        if field_name not in json_schema["required"]:
+            json_schema["required"].append(field_name)
 
     if multi:
         return multi_schema(json_schema)
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -226,6 +226,32 @@ def test_extract_fenced_code_block(input, last, expected):
                 "required": ["name", "age"],
             },
         ),
+        # Test case 9: Repeated property name is listed once in required
+        (
+            "name, age int, name",
+            {
+                "type": "object",
+                "properties": {
+                    "name": {"type": "string"},
+                    "age": {"type": "integer"},
+                },
+                "required": ["name", "age"],
+            },
+        ),
+        # Test case 10: Repeated name keeps the last description
+        (
+            """
+        name: first name
+        name: last name
+        """,
+            {
+                "type": "object",
+                "properties": {
+                    "name": {"type": "string", "description": "last name"},
+                },
+                "required": ["name"],
+            },
+        ),
     ],
 )
 def test_schema_dsl(schema, expected):


simonw:
I made this raise `ValueError` instead: https://github.com/simonw/llm/commit/f947edd5d7d249442807a80a53ca60f11b860456

PR #1620, opened 2026-08-18
Add --key option to llm embed and llm embed-multi

### LINKED ISSUE #757: llm embed --key option
`llm embed` should take a `--key` option for consistency with `llm prompt`.

ChrisJr404:
Closes #757.

`llm prompt` already accepts a `--key` option to pass an API key directly. This adds the same option to `llm embed` and `llm embed-multi` so an embedding model's key can be supplied on the command line rather than only via a stored key (`llm keys set ...`) or an environment variable.

When `--key` is given, it is set on the resolved embedding model before embedding, so `EmbeddingModel.get_key()` returns it (mirroring how the key is threaded through for prompts). The default behaviour is unchanged when the option is omitted.

Example:

```bash
llm embed -c 'This is some content' -m 3-small --key sk-...
```

### Changes
- `--key` option on both `embed` and `embed-multi` in `llm/cli.py`
- Docs: mention in `docs/embeddings/cli.md`; regenerated `docs/help.md` via cog
- Tests: `test_embed_key_option` and `test_embed_multi_key_option` verifying the key reaches the model

`pytest tests/test_embed_cli.py tests/test_embed.py` passes (109 passed); `black` and `ruff` are clean.

### DIFF
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -3261,6 +3261,7 @@ def uninstall(packages, yes):
 @click.option(
     "-m", "--model", help="Embedding model to use", envvar="LLM_EMBEDDING_MODEL"
 )
+@click.option("--key", help="API key to use")
 @click.option("--store", is_flag=True, help="Store the text itself in the database")
 @click.option(
     "-d",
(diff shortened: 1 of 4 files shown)


llm/cli.py
```diff
@@ -3498,6 +3514,9 @@ def embed_multi(
             "You need to specify an embedding model (no default model is set)"
         )
 
+    if key:
+        collection_obj.model().key = key
```
simonw, line 3518: This pattern feels uncomfortable to me - it relies on `.model()` returning a cached value.

I'm inclined to accept the rest of this change though, then I'm going to refactor this to a cleaner pattern.

simonw: See e215c421127ac9aa696f4d94a563e80169029e29

PR #1622, opened 2026-08-20
Declare typing-extensions, and a test that catches undeclared imports

vanshtaneja23:
Refs #1608. Follow-on to e9a4382 — that fixed `httpx`; this covers the same
class of bug that is still latent, and adds a test so there isn't a third one.

`llm/serialization.py` imports `typing_extensions` directly, and the comment
above the import says out loud that it is leaning on a transitive:

```python
# NotRequired moved to typing in 3.11; use typing_extensions for 3.10
# support. typing_extensions is a transitive dep via pydantic.
from typing_extensions import NotRequired, TypedDict
```

That is exactly the arrangement `httpx` had via `openai` until openai 3.0
switched to httpx2 and every fresh install started dying at import time. It
works today because pydantic pulls it in, and it keeps working right up until
pydantic stops — at which point it breaks the same way, on users' machines
rather than in CI. The import is unconditional, so it is needed on every
supported Python, not just 3.10.

### Changes

**`pyproject.toml`** — add `typing-extensions` to `dependencies`.

**`tests/test_llm.py`** — new `test_all_third_party_imports_are_declared`:
walks the AST of every module under `llm/`, collects top-level non-stdlib
imports, maps each to its providing distribution via
`importlib.metadata.packages_distributions()`, and asserts it appears in the
package's own declared requirements.

On current `main` it fails with:

```
AssertionError: these modules are imported by llm/ but are not declared in
pyproject.toml dependencies: typing_extensions
```

and passes with the `pyproject.toml` change. It generalises, so the next
undeclared import gets caught in CI instead of by a user on a fresh install.

### Verification

- Verified failing before the `pyproject.toml` change, passing after.
- `typing_extensions` is the only thing it flags — no false positives across
  the rest of `llm/`.
- `black --check .`, `ruff check .` and `mypy llm` are clean.

### Notes on the test

- Reads sources with an explicit `encoding="utf-8"` rather than the platform
  default, since files under `llm/` contain non-ASCII — otherwise it would
  fail on the Windows CI matrix.
- Skips relative imports, and skips modules it cannot resolve to an installed
  distribution rather than guessing, so an optional/platform-gated import can't
  produce a spurious failure.
- Ignores requirements gated behind an `extra ==` marker, so only genuinely
  guaranteed dependencies count as declared.

I have deliberately left the second half of #1608 (the suite escaping to the
live API under openai 3) alone — #1621 is taking that on properly, and this
does not touch it. Note that `Test` on `main` is currently red for that reason,
so CI on this PR will likely be red too; it is pre-existing and unrelated to
this change.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,6 +32,7 @@ dependencies = [
     "click-default-group>=1.2.3",
     "sqlite-utils>=4.0",
     "pydantic>=2.0.0",
+    "typing-extensions",
     "PyYAML",
     "pluggy",
     "python-ulid",
--- a/tests/test_llm.py
+++ b/tests/test_llm.py
@@ -1,6 +1,10 @@
+import ast
 import json
 import os
 import pathlib
+import re
+import sys
+from importlib import metadata
 from importlib.metadata import version
 from unittest import mock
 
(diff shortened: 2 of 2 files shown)


simonw:
Adding a test to avoid repeats of the `httpx` incident is a good idea, thanks.

PR #1623, opened 2026-08-20
Fix --data-ids attaching wrong conversation_id to log items (#1598)

### LINKED ISSUE #1598: --data-ids attaches wrong conversation_id (uses response id instead)

pri5h:
## Summary
This PR fixes a bug where `llm logs --data --data-ids` was incorrectly setting the `conversation_id` to the individual response ID instead of the actual conversation ID from the log row.

## Changes
- **Fix**: Updated `llm/cli.py` to use `row["conversation_id"]` instead of `row["id"]` when attaching the `conversation_id` to data items.
- **Test**: Added a regression test in `tests/test_llm_logs.py` to ensure `response_id` and `conversation_id` remain distinct, preventing this bug from recurring.

Closes #1598

### DIFF
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -150,9 +150,7 @@ def _run_chat(
     click.echo("Type '!multi' to enter multiple lines, then '!end' to finish")
     click.echo("Type '!edit' to open your default editor and modify the prompt")
     if db is not None:
-        click.echo(
-            "Type '!fragment <my_fragment> [<another_fragment> ...]' to insert one or more fragments"
-        )
+        click.echo("Type '!fragment <my_fragment> [<another_fragment> ...]' to insert one or more fragments")
 
     argument_fragments = list(initial_fragments or [])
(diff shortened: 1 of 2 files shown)


simonw:
Fixed in:
- #1613

PR #1628, opened 2026-08-21
Validate attachments in conversation prompts

### LINKED ISSUE #1626: Conversation.prompt() skips attachment validation that Model.prompt() enforces
`Model.prompt()`, `AsyncModel.prompt()`, `Conversation.chain()` and `AsyncConversation.chain()` all call `self.model._validate_attachments(attachments)`, but `Conversation.prompt()` and `AsyncConversation.prompt()` do not (checked against llm 0.32, `llm/models.py` — sync `Conversation.prompt` around line 822 builds the `Prompt`/`Response` without the validation call the other entry points have).

Since the CLI routes continued conversations through `conversation.prompt` (cli.py `prompt_method = conversation.prompt`), the practical effect: the *first* `llm -m model ... -a file` is correctly rejected client-side, but the same attachment on a follow-up `llm -c ... -a file` (or inside `llm chat`) is sent to the provider. For hosted models that's a paid API error instead of a free local one.

Repro (no network needed):

```python
import llm
from llm.models import Attachment

class Fake(llm.Model):
    model_id = "fake"
    attachment_types = set()  # text-only
    def execute(self, prompt, stream, response, conversation=None):
        yield "SENT"  # reaching here means the request would have gone out

m = Fake()
att = Attachment(type="image/png", content=b"px")

m.prompt("hi", attachments=[att]).text()                 # ValueError: This model does not support attachments
m.conversation().prompt("hi", attachments=[att]).text()  # prints nothing, returns "SENT"
```

Output on 0.32:

```
Model.prompt: blocked -> This model does not support attachments
Conversation.prompt: SENT (no validation!)
```

Suggested fix: add the same `self.model._validate_attachments(attachments)` call at the top of `Conversation.prompt()` and `AsyncConversation.prompt()`, matching `chain()`.

Found while building [llm-tiyuvta](https://github.com/avifenesh/llm-tiyuvta); the plugin now guards at `execute()` time as a workaround. Happy to send a PR if useful.

original4422:
## Summary

- validate attachments in sync and async `Conversation.prompt()` before a response can execute the provider
- cover both models with no attachment support and models that reject a specific MIME type
- verify supported attachments and the resulting multi-turn conversation history remain unchanged

## Root cause

`Model.prompt()`, `AsyncModel.prompt()`, and both conversation `chain()` paths already call the model attachment validator. The two conversation `prompt()` entry points skipped that same contract, so continued CLI conversations could send an unsupported attachment to the provider.

## Validation

- `pytest -q tests/test_attachments.py` — 13 passed
- `pytest -q` — 1,099 passed with the current `httpx2` / OpenAI 3 dependency set
- `black --check .`
- `ruff check .`
- `mypy llm`
- Cog documentation check
- `python -m build`
- `tests/test-llm-load-plugins.sh` with `llm-cluster` and `llm-mistral`

The branch has been merged normally with current upstream `main` (`7008f55`), which includes the upstream HTTP/OpenAI test-compatibility update. No dependency change is part of this PR's diff.

Fixes #1626

### DIFF
--- a/llm/models.py
+++ b/llm/models.py
@@ -837,6 +837,7 @@ class Conversation(_BaseConversation):
         hide_reasoning: bool = False,
         **kwargs,
     ) -> "Response":
+        self.model._validate_attachments(attachments)
         merged = _merge_options(options, kwargs)
         # Build the authoritative chain so response.prompt.messages
         # equals exactly what the model sees for this turn.
@@ -1023,6 +1024,7 @@ class AsyncConversation(_BaseConversation):
         hide_reasoning: bool = False,
         **kwargs,
     ) -> "AsyncResponse":
+        self.model._validate_attachments(attachments)
         merged = _merge_options(options, kwargs)
         chain = self._build_full_chain(
             prompt=prompt,
--- a/tests/test_attachments.py
+++ b/tests/test_attachments.py
@@ -159,3 +159,108 @@ def test_attachment_resolve_type_limits_redirects(httpx_mock):
         attachment.resolve_type()
 
     assert len(httpx_mock.get_requests()) == 4
+
+
+UNSUPPORTED_ATTACHMENT_CASES = (
+    ((), "This model does not support attachments"),
+    (
+        ("audio/wav",),
+        "This model does not support attachments of type 'image/png', only audio/wav",
+    ),
+)
+
+
+@pytest.mark.parametrize(
+    "attachment_types,expected_message", UNSUPPORTED_ATTACHMENT_CASES
+)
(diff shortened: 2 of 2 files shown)


original4422:
Synced this branch with current upstream `main` in `93447e8` using an ordinary merge (no history rewrite). This includes the upstream HTTP/OpenAI test-compatibility update that the unchanged old-base CI run was missing.

Validation on the merged head:

- conversation attachment regression cases: 6 passed
- complete attachment tests: 13 passed
- complete suite: 1,099 passed
- Black, Ruff, mypy, and Cog checks: passed
- sdist/wheel build and plugin-load check: passed

CI has started for exact head `93447e8`.

PR #1641, opened 2026-08-24
Keep monotonic_ulid() monotonic when the clock does not move forward

dylanpulver:
`monotonic_ulid()` promises a ULID strictly larger than every other from the same process, and claims parity with the reference JavaScript `monotonicFactory`. The increment branch fires only on `now_ms == last_ms` (`llm/utils.py:746`), so a reading at or below the previous timestamp falls through to `_fresh()` and yields a smaller ULID.

Two ways that reading arrives:

- The wall clock steps backwards: `time.time_ns()` is not monotonic, so an NTP step or VM resume moves it.
- `time.time_ns()` is read outside `_lock`, so a thread holding the earlier millisecond can reach the lock after a thread holding the later one. The lock is there because concurrent callers are in scope.

These IDs become conversation and turn ids in `logs.db`, where the ULID sorts rows by creation.

The fix tests `now_ms <= last_ms`, matching `ulid/javascript` `source/ulid.ts:161` (`if (seed <= lastTime)`), which likewise reuses the previous timestamp rather than encoding the regressed one. Moving `time.time_ns()` inside the lock closes the thread window but not clock regression.

Residual: the guarantee stays per-process, and a regressed clock now draws randomness from the pinned millisecond, so a long regression reaches the existing `OverflowError`.

1107 tests pass locally against 1105 on main. Both new tests fail on `==` and pass on `<=`; `test_monotonic_ulids` passes either way. Black, ruff, mypy and cog are clean. No changelog entry, since `docs/contributing.md` assigns that to the release process.

### DIFF
--- a/llm/utils.py
+++ b/llm/utils.py
@@ -722,8 +722,9 @@ def monotonic_ulid() -> ULID:
     other ULID returned by this function inside the same process.
 
     It works the same way the reference JavaScript `monotonicFactory` does:
-    * If the current call happens in the same millisecond as the previous
-        one, the 80-bit randomness part is incremented by exactly one.
+    * If the current call reads a millisecond that is not later than the
+        previous one, the 80-bit randomness part is incremented by exactly
+        one and the previous timestamp is reused.
     * As soon as the system clock moves forward, a brand-new ULID with
         cryptographically secure randomness is generated.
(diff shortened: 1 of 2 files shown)


simonw:
Older PRs:
- #1490
- #1454

PR #1644, opened 2026-08-27
Fix fenced code extraction with CRLF line endings

mameikagou:
Fenced model responses using CRLF line endings were not recognized. `--extract` could therefore print the entire reply or skip to a later LF-only code block.

Accept CRLF at opening and closing fences while preserving the extracted content, language-tag syntax, and fence-length checks. Add utility and mocked-HTTP CLI regressions, including mixed line endings and first/last selection.

Validation on Python 3.14:

- `python -m pytest -q`: 1109 passed.
- `black --check .`, `mypy llm`, and `ruff check .`: passed.
- The new CLI regression fails against the unchanged upstream function and passes with this fix.


simonw:
Thanks for this.

PR #1651, opened 2026-09-01
cache generated options classes to speed up model lookup

waveplate:
## problem

`Responses` and `AsyncResponses` currently call `build_options_class()` for every model instance. this creates a new pydantic model class each time, even when another model has exactly the same capabilities.

this becomes expensive when a plugin registers a large number of models, since the same options classes are rebuilt whenever `llm.get_models()` or `llm.get_model()` is called.

## change

this caches `build_options_class()` using its existing capability flags. different capability combinations still receive separate classes, while identical combinations reuse the class that has already been built.

## timings

| llm version | llm-openrouter version | repeated `llm.get_models()` |
|---|---|---:|
| `llm==0.31.1` | `llm-openrouter==0.6.0` | 0.435 seconds |
| `llm==0.33` | `llm-openrouter==0.6.0` | 0.908 seconds |
| `llm==0.33` | `llm-openrouter==0.7` | 18.9 seconds |
| `llm==0.33` with this change | `llm-openrouter==0.7` | 5.531 seconds |
| `llm==0.33` with this change | `llm-openrouter==0.7` with its corresponding cache | 0.027 seconds |

## tests

- 1,106 tests passed
- black passed
- ruff passed
- mypy passed
- documentation checks passed

### DIFF
--- a/llm/default_plugins/openai_models.py
+++ b/llm/default_plugins/openai_models.py
@@ -4,6 +4,7 @@ import os
 import sys
 from collections.abc import AsyncGenerator, Iterable, Iterator
 from enum import Enum
+from functools import cache
 from typing import Any, ClassVar, Literal
 
 import click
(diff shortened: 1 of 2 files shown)


simonw:
Related:
- #1654

simonw:
This is a neat, simple fix. Thanks.

Issue #1654, opened 2026-09-02
Performance issues with llm logs


simonw:
`llm logs` can be VERY slow for conversations with a lot of messages in them, e.g. agent tool sessions.

simonw:
Here's a Codex investigation into this. My summary:

- Every log row resolves models, which involve a call to `register_models()`
- If you have `llm-openrouter` installed it iterates through and inflates objects for 400+ models each time it is called
- Workaround now is to cut down the plugins: `LLM_LOAD_PLUGINS=llm-gemini llm logs -cue`

<details><summary>Codex report</summary>

# `llm logs -c` repeatedly rebuilds the model registry for every log row

## Summary

`llm logs -cue` becomes very slow for conversations with many turns when
`llm-openrouter` is installed. The slowdown is caused by `llm` resolving the
model separately for every log row while rehydrating `response_json`.

Each model lookup invokes every plugin's `register_models` hook. The
`llm-openrouter` hook currently constructs synchronous and asynchronous model
objects for every model returned by OpenRouter, including fresh Pydantic option
classes. This work is repeated once per turn even when the logged conversation
does not use OpenRouter.

The `-u` and `-e` options are not responsible for the slowdown. `-c` is the
important option because it changes the default limit from three rows to every
turn in the current conversation.

## Environment

- `llm` 0.33
- Executable: `~/.local/bin/llm`, installed as a uv tool
- macOS
- Relevant installed plugins:
  - `llm-openrouter` 0.7
  - `llm-gemini` 0.34
  - `llm-anthropic` 0.27
  - `llm-lmstudio` 0.3.1
  - `llm-meta-ai` 0.1
  - `llm-coding-agent` 0.1a0
  - `llm-jq` 0.1.1
  - `llm-chat-completions-server` 0.1a0

## Reproduction

Use `llm` to create a conversation with a substantial number of agent or tool
turns, then run:

```bash
/usr/bin/time -p llm logs -cue >/dev/null
```

To compare a bounded number of rows with and without OpenRouter:

```bash
/usr/bin/time -p llm logs -cue -n 3 >/dev/null

LLM_LOAD_PLUGINS='llm-lmstudio,llm-meta-ai,llm-anthropic,llm-coding-agent,llm-gemini,llm-jq,llm-chat-completions-server' \
  /usr/bin/time -p llm logs -cue -n 3 >/dev/null
```

## Observed behavior

The original full command took 38.85 seconds when the current conversation had
51 turns.

[...]

PR #1655, opened 2026-09-02
Cache build_options_class() - model registration rebuilt pydantic classes per instance

rdslw:
`build_options_class()` is a pure function of five booleans (at most 32 distinct classes), but every OpenAI-compatible model instance called `pydantic.create_model()` afresh. A plugin registering many models pays for it on every `register_models` pass, and `get_model()` re-runs that pass on each call. With llm-openrouter installed (~1200 models × 2 classes) one pass took ~7.5s.

This caches the class per flag combination with `functools.cache`. Tests check identity per combination and the field shape.

Wall clock with llm-openrouter, llm-gemini, llm-cerebras and llm-openai-codex installed (hyperfine, 3 runs):

| Command | 0.33 | 0.33 + this PR |
|---|---|---|
| `llm models list` | 11.0 s | 2.3 s |
| `llm logs list -m cerebras-gpt-oss-120b -n 5` | 18.7 s | 2.9 s |
| `llm logs list -n 0` (361 rows) | 37.6 s | 3.7 s |

The #1654 optimizations skip `get_model()` only for rows without model markers; rows that carry them still pay a full registration per model, which this fixes at the source.

Created with help of Fable-5.1 during migration of llm-openai-codex from llm-0.32 to llm-0.33.

### DIFF
--- a/llm/default_plugins/openai_models.py
+++ b/llm/default_plugins/openai_models.py
@@ -1,4 +1,5 @@
 import datetime
+import functools
 import json
 import os
 import sys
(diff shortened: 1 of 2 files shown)


rdslw:
Superseded by #1651, which landed the same `@cache` on `build_options_class()` shortly after this was opened. Closing.

PR #1659, opened 2026-09-03
Close provider generator when async event stream closes

imshunsuke:
When `AsyncResponse.astream_events()` is explicitly closed or finalized before normal exhaustion, it currently leaves the underlying provider async generator open. That can delay provider cleanup for streaming HTTP responses and plugin-managed resources.

Propagate early closure to the provider generator while preserving the existing normal-completion path through `_async_finalize()`. Add a regression test that consumes one event through `contextlib.aclosing` and verifies that provider cleanup runs.

Validation:

- `uv run pytest`: 1125 passed
- `uv run black . --check`: passed
- `uv run mypy llm`: passed
- `uv run ruff check .`: passed
- package sdist and wheel build: passed
- the new regression fails on the untouched base and passes with this change

The Cog generated-fragment check still reports the existing `docs/fragments.md (changed)` result, which was reproduced on the untouched base commit.

### DIFF
--- a/llm/models.py
+++ b/llm/models.py
@@ -2561,7 +2561,8 @@ class AsyncResponse(_BaseResponse):
                 self._process_chunk(chunk)
                 yield self._stream_events[-1]
         finally:
-            pass
+            if not self._done:
+                await self._generator.aclose()
 
     async def messages(self) -> list[Any]:
         """List of Message objects produced by this response.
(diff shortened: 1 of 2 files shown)


anandfresh (approved):

simonw:
Thanks for catching this.

PR #1665, opened 2026-09-05
Fix `llm -- prompt` not treating argument as positional

### LINKED ISSUE #245: escaping double dashs
```powershell
(base) PS C:\Users\user> llm "--find me"
Usage: llm prompt [OPTIONS] [PROMPT]
Try 'llm prompt --help' for help.

Error: No such option: --find me
(base) PS C:\Users\user> llm '--find me'
Usage: llm prompt [OPTIONS] [PROMPT]
Try 'llm prompt --help' for help.

Error: No such option: --find me
(base) PS C:\Users\user>
```

ikatyal2110:
When running `llm -- "--find me"`, the cli group (DefaultGroup) consumes the `--` separator and forwards `"--find me"` to the prompt subcommand without the separator, so Click's parser mistakes it for an option and raises "No such option: --find me". The equivalent `llm prompt -- "--find me"` works correctly because the separator reaches the prompt command intact.

The fix detects when `--` is the first argument to the cli group and injects the default command name (`prompt`) before it, so `['--', arg]` becomes `['prompt', '--', arg]` and the separator is forwarded to the subcommand as intended. A regression test using the echo model verifies the prompt value is received correctly.

Fixes #245

---
_Generated by [Claude Code](https://claude.ai/code/session_015J3gVny9MN8kFEQaQvMqSk)_

### DIFF
--- a/llm/cli.py
+++ b/llm/cli.py
@@ -24,6 +24,19 @@ import pydantic
 import sqlite_utils
 import yaml
 from click_default_group import DefaultGroup
+
+
+class _DefaultGroupWithDoubleDash(DefaultGroup):
+    def parse_args(self, ctx, args):
+        # When the user writes `llm -- <arg>` the group's parser consumes `--`
+        # and passes `<arg>` to the subcommand resolver without the separator.
+        # If `<arg>` starts with `-` the subcommand's parser then mistakes it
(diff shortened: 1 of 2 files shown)


simonw:
See also the upstream bug I filed:
- https://github.com/click-contrib/click-default-group/issues/25

PR #1670, opened 2026-09-06
Fix pipx installation link

ssmurfgg04-gif:
## What
The pipx install instruction points to pypa.github.io/pipx/, which returns 404. Updated it to pipx.pypa.io, the current docs site.

## Why
The pipx docs moved off pypa.github.io, the old path is dead.

## Verification
Old URL 404, new URL 200.


simonw:
Thanks!

simonw:
I also filed this bug:
- https://github.com/pypa/pipx/issues/2030

ssmurfgg04-gif:
You're welcome! And thanks for filing that pipx bug — I hadn't realized the broken link traced back to the pipx side rather than just docs drift. I'll keep an eye on it; that same `pypa.github.io/pipx` link is probably still live in plenty of other projects' install docs, so good to have it tracked upstream.

PR #1683, opened 2026-09-14
Fix dead Jina embeddings documentation links

auroraxo:
## Summary

Replace two references to a removed Jina AI announcement with the canonical Hugging Face model card for `jina-embeddings-v2-base-en`.

The old URL now returns HTTP 404; the model card returns HTTP 200 and preserves the specific 8K-context model information those sentences cite.

## Verification

- Confirmed both old references return 404
- Confirmed the replacement URL returns 200
- Documentation-only change; no runtime behavior changed


auroraxo:
Context on how I found these: I ran a documentation link-rot scan over `simonw/llm@main` (31 docs files, 831 outbound links checked) and reviewed the results by hand before opening this PR.

Two other findings from that scan that I did **not** touch here, because they need a maintainer decision rather than a mechanical fix:

- `docs/logging.md:438` — the BLAKE2b link is `https://en.wikipedia.org/wiki/BLAKE_(hash_function` (missing the closing paren, so it 404s). The intended target `https://en.wikipedia.org/wiki/BLAKE_(hash_function)` returns 200. Markdown's handling of the trailing `)` is the reason it got truncated; `<...>` angle brackets fix it.
- `docs/fragments.md:100` — the sample output references `https://raw.githubusercontent.com/simonw/llm-docs/refs/heads/main/llm/0.22.txt`, which is now 404 (the `llm/` tree is gone from `simonw/llm-docs`). It is illustrative output rather than a live link, so it may be intentional.

Happy to fold either of those into this PR if you'd like them fixed the same way.

anandfresh (approved):

PR #1686, opened 2026-09-15
Fix supports_schema attribute name in advanced model plugins docs

### LINKED ISSUE #814: Small syntax error in the schema docs
I think this needs to be `supports_schema`.

https://github.com/simonw/llm/blob/e060347f58fd281c75f407745c4814d767f83d8c/docs/plugins/advanced-model-plugins.md?plain=1#L102

ethanstoner:
Closes #814

The schema example in `docs/plugins/advanced-model-plugins.md` sets `support_schema = True`, but the attribute llm checks (`supports_schema` in `llm/models.py`, and the sentence just above the example) has an `s`. A plugin author copying the snippet would silently get no schema support.

AI disclosure: this change was made with Claude Code; I checked the attribute name against `llm/models.py`.


anandfresh (approved):

simonw:
Thanks!

PR #1694, opened 2026-09-20
Fix $(O) typo in docs Makefile livehtml target

v1teka:
The `livehtml` target in `docs/Makefile` passes `$(0)` where every other target passes `$(O)`. Make expands `$(0)` to an empty string, so options given the documented Sphinx way are silently dropped:

```
$ make -n livehtml O="--port 9000"
sphinx-autobuild -b html "." "_build"
```

With the fix:

```
$ make -n livehtml O="--port 9000"
sphinx-autobuild -b html "." "_build" --port 9000
```

`just docs` passes the port through `SPHINXOPTS`, so it behaves the same before and after.

This also updates `SPHINXPROJ`, which still held the `sqlite-utils` value this Makefile was originally copied from. It is unused by the current Sphinx Makefile, but the stale name is confusing next to `project = "LLM"` in `conf.py`.


anandfresh (approved):

v1teka:
Gentle ping — no rush. Two-line docs/Makefile fix; happy to narrow it to just
the $(O) typo if you'd prefer a smaller diff. @simonw

